Issue #4, opened 2014-01-05
message-bus.js: Exception on Connection Timeout


felixbuenemann:
There is an unhandled error case in the message-bus.js client:

If the long polling connection to the server times out without receiving the default `[]` reply, eg. because the network connection times out earlier, the following error is thrown:

```
[Error] TypeError: 'null' is not an object (evaluating 'object.length')
    each (jquery.js, line 633)
    success (message-bus.js, line 123)
    resolveWith (jquery.js, line 1017)
    done (jquery.js, line 7248)
    callback (jquery.js, line 8029)
```

This is because in this case the content of messages in message-bus.js:123 is not an array, so the `each` implementation in jQuery fails to call `length` on it. Checking for `messages == null` should fix this. I'll make a PR to fix this.

I noticed this, because Thin has a connection timeout of 30 seconds and the default `MessageBus.long_polling_interval` is also 30 seconds. So there is a race condition between those two timeouts. I would also recommend to change the default polling interval to 29 seconds or maybe 25 seconds (that's what Faye uses in all examples).

PR #6, opened 2014-01-05
Change default long polling interval to 29 seconds

felixbuenemann:
This is related to issue #4: When running on Thin, connections time out
after 30 seconds. So in order to avoid race conditions, we need to send
our reply to the client slightly earlier.
In most other server setups you should be fine with 59 seconds.

### DIFF
--- a/lib/message_bus.rb
+++ b/lib/message_bus.rb
@@ -88,7 +88,7 @@ module MessageBus::Implementation
   end
 
   def long_polling_interval
-    @long_polling_interval || 30 * 1000
+    @long_polling_interval || 29 * 1000
   end
 
   def off
--- a/spec/lib/middleware_spec.rb
+++ b/spec/lib/middleware_spec.rb
@@ -59,7 +59,7 @@ describe MessageBus::Rack::Middleware do
         MessageBus.long_polling_interval = 10
         s = Time.now.to_f * 1000
         post "/message-bus/ABC", '/foo' => nil
-        (Time.now.to_f * 1000 - s).should < 30
+        (Time.now.to_f * 1000 - s).should < 29
       ensure
         MessageBus.long_polling_interval = 5000
       end


SamSaffron:
fyi I just added streaming support, try it out if you are still using message_bus

felixbuenemann:
I'm currently using faye and booting up a pair of thins behind nginx with a dedicated rackup config to handle pubsub messaging. This allows to immediately failover clients during deploys using rolling restarts. It works, but the setup and monitoring is much more complex than I'd like.

I still have the message_bus code in place swappable by configuration, so I could try it out.

In addition to broadcasted notifications I'm also pushing status updates and flash messages to clients which would benefit from the reduced latency/connection overhead of the new streaming support, so it makes sense to retest message_bus against faye.

@SamSaffron I'm still concerned about concurrency. The README mentions that it's not an issue, due to the use of rack hijack, but this depends heavily on how the rack server implements hijacking. I'm currently using phusion passenger as an app server and it does not accept new connections after a request is highjacked, so on passenger each polling client will block a worker process until the polling connection is closed, which means running the middleware inside the app is a bad idea (tested with passenger standalone 5.0.23).

I also tested unicorn, thin and puma using `examples/bench/config.ru` which have different behavior: Normal app requests are unaffected, but only one long polling client can be connected. As soon as the second client connects, the first polling connection is closed. The ab bench example also shows that only  one request has the duration of the max polling interval, while all other finish in milliseconds:
```
Percentage of the requests served within a certain time (ms)
  50%     19
  66%     21
  75%     22
  80%     23
  90%     25
  95%     27
  98%     33
  99%     38
 100%   2009 (longest request)
```
I would have expected all requests to take roughly 2 seconds (the configured `long_polling_interval`).

The behavior on passenger is probably due to it's architecture where a c++ agent talks to the preforked workers, but the behavior on the other servers seem faulty pn both 1.1.1 and master. **Should I open a new issue for that?**

SamSaffron:
ahh that bench is totally faulty got to fix it but can no longer use ab to
run it.

I added protection into message_bus to ensure only one client_id is handled
at a time so if you want to test concurrency you need to generate unique
client ids for the long polls, this is a pain in the behind to do in ab,
maybe worth writing a simple ruby client for testing this. If duplicate
reqs come in from the same client ids only the latest wins and earlier ones
are cancelled (which is the correct behavior, we use _seq to ensure we are
sequenced on latest)

passenger issue is very concerning we should raise it with @foobarwidget if
it is reproducible.

I think the first thing to do here is write a correct bench that simulates
multiple clients talking to the server.

SamSaffron:
Note, nothing is stopping you running message bus dedicated if you want
to... but actual resource usage on it is quite minimal unless you have tons
of load.

felixbuenemann:
wg/wrk supports lua scripts, I'll look at the required headers and see if I can come up with a proper benchmark script.

SamSaffron:
awesome, the key is having unique client ids for every req.

Let me know how you go!

PR #27, opened 2015-04-21
Quick n' dirty auto-scroll

tms:
WIth <3


SamSaffron:
thanks :+1: 

about to be deployed

PR #29, opened 2015-05-11
use max_backlog_size parameter

jraedisch:
Actually use max_backlog_size parameter (that is already provided) when initializing ReliablePubSub.


SamSaffron:
oops, good catch !

PR #30, opened 2015-05-11
allow configuring max_backlog_size via MessageBus::Implementation

jraedisch:


SamSaffron:
I really want us to have reliable pub sub implementations for both pg and memory. I wonder if this interface is right. 

There are 2 backlogs, one per channel and one global one, only passing in the per-channel one is a bit weird and I wonder if it even should go in via the initializer (if it does it mean we should have both)

jraedisch:
You are probably right. The fix solves my problem but there might be usecases where people want to change other limits, and putting all of them into the API is too much.

What about a setter for reliable_pub_sub then?
What other solution would you suggest?

SamSaffron:
A setter in reliable pub sub is fine, we should prob rename to
MessageBus::Redis::ReliablePubSub  as prep work

jraedisch:
Sorry, with setter I meant a setter for reliable_pub_sub in MessageBus::Implementation, so I can provide one with custom limits. 
This would theoretically even allow overriding redis dependency. (Ah, and that is why you proposed moving it to MessageBus::Redis?!)
Alternative would be for each user to extend MessageBus by themselves and override reliable_pub_sub getter, I think.

PR #32, opened 2015-05-21
allow providing xhrFields like 'withCredentials' for CORS

jraedisch:
When connecting to long-polled server on a different domain/port you might want to share authorization cookies with it, so you have to add xhrFields { withCredentials: true } to ajax calls.
This patch would enable doing this.


SamSaffron:
something feels a bit odd here, maybe a cleaner way of just adding ANY options to the xhr?

jraedisch:
Makes sense. I will prepare another pull request.

PR #33, opened 2015-05-26
allow overriding all ajax settings

jraedisch:
Like this, all ajax settings can be overridden. 
Not sure, if this should be limited somehow, or if performance could be tweaked, but it works.

    MessageBus.ajaxOverrides = { xhrFields: { withCredentials: true }};


SamSaffron:
I am not sure we need this ... MessageBus.ajax is already exposed so you can simply wire up a different method there and do all the overriding you want. 

function myAjax(opts) {
    opts.something = "bla";
    return $.ajax(opts)
}

MessageBus.ajax = myAjax; 

I think the correct course of action here is simply to document this.

jraedisch:
Ah, I didn't see that! Works.

PR #34, opened 2015-06-04
allow overriding longPoller

jraedisch:
Now I wanted to supply some more data to the backend, but couldn't, since longPoller wasn't exposed. This would help. (Or is there a better way to add properties to data?)

When adding data this way, you have to make sure to delete it from env['rack.request.form_hash'] before passing env on to MessageBus middleware - otherwise it can lead to errors when MessageBus tries to convert that data to channels.

### DIFF
--- a/assets/message-bus.js
+++ b/assets/message-bus.js
@@ -162,6 +162,7 @@ window.MessageBus = (function() {
     callbacks: callbacks,
     clientId: clientId,
     alwaysLongPoll: false,
+    longPoller: longPoller,
     baseUrl: baseUrl,
     // TODO we can make the dependency on $ and jQuery conditional
     // all we really need is an implementation of ajax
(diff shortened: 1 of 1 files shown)


SamSaffron:
can you send a full snippet of what you are trying to do?

jraedisch:
I am using message_bus for pseudo real-time updating of charts. Since I do not want do get costly 3rd party data for charts, that are not currently opened, I am sending a chart id with every message bus request like this (slightly simplified):

        MessageBus.originalLongPoller = MessageBus.longPoller;

        MessageBus.longPoller = function(poll,data){
            data.chart = {id: myLib.chart.id};
            MessageBus.originalLongPoller(poll,data);
        }

In the backend I have a middleware before message_bus middleware, that deletes the chart:

    def call(env)
        return @app.call(env) unless env['PATH_INFO'] =~ /^\/message-bus\//
        chart_hash = env['rack.request.form_hash'].delete('chart')
        redis.setex(chart['id'], Constants.long_polling_timeout, 'true')
        @app.call(env)
    end

SamSaffron:
I am going to close this since I ensured $.ajax is configurable, you can plug in a custom ajax implementation to take care of this

Issue #24, opened 2015-03-05
ActionDispatch::Flash shouldn't be required


arcreative:
IMO, `app.middleware.insert_before(ActionDispatch::Flash, MessageBus::Rack::Middleware)` is a bit naïve since you don't know that the project will have it required.  In my rails-api project, it errors out: 

```
=> Booting WEBrick
=> Rails 4.1.8 application starting in development on http://0.0.0.0:3000
=> Run `rails server -h` for more startup options
=> Notice: server is listening on all interfaces (0.0.0.0). Consider using 127.0.0.1 (--binding option)
=> Ctrl-C to shutdown server
Exiting
/Users/adam/.rvm/gems/ruby-2.2.0/gems/actionpack-4.1.8/lib/action_dispatch/middleware/stack.rb:125:in `assert_index': No such middleware to insert before: ActionDispatch::Flash (RuntimeError)
	from /Users/adam/.rvm/gems/ruby-2.2.0/gems/actionpack-4.1.8/lib/action_dispatch/middleware/stack.rb:88:in `insert'
	from /Users/adam/.rvm/gems/ruby-2.2.0/gems/railties-4.1.8/lib/rails/configuration.rb:68:in `block in merge_into'
	from /Users/adam/.rvm/gems/ruby-2.2.0/gems/railties-4.1.8/lib/rails/configuration.rb:67:in `each'
	from /Users/adam/.rvm/gems/ruby-2.2.0/gems/railties-4.1.8/lib/rails/configuration.rb:67:in `merge_into'
	from /Users/adam/.rvm/gems/ruby-2.2.0/gems/railties-4.1.8/lib/rails/engine.rb:497:in `app'
	from /Users/adam/.rvm/gems/ruby-2.2.0/gems/railties-4.1.8/lib/rails/application/finisher.rb:36:in `block in <module:Finisher>'
	from /Users/adam/.rvm/gems/ruby-2.2.0/gems/railties-4.1.8/lib/rails/initializable.rb:30:in `instance_exec'
	from /Users/adam/.rvm/gems/ruby-2.2.0/gems/railties-4.1.8/lib/rails/initializable.rb:30:in `run'
	from /Users/adam/.rvm/gems/ruby-2.2.0/gems/railties-4.1.8/lib/rails/initializable.rb:55:in `block in run_initializers'
	from /Users/adam/.rvm/rubies/ruby-2.2.0/lib/ruby/2.2.0/tsort.rb:226:in `block in tsort_each'
[...]

PR #35, opened 2015-06-08
Append message_bus to the end of the middleware stack.

tgxworld:
fixes https://github.com/SamSaffron/message_bus/issues/24.

### DIFF
--- a/lib/message_bus/rails/railtie.rb
+++ b/lib/message_bus/rails/railtie.rb
@@ -11,9 +11,9 @@ class MessageBus::Rails::Railtie < ::Rails::Railtie
     # the Rails app is configured that might be ActionDispatch::Session::CookieStore, or potentially
     # ActionDispatch::Session::ActiveRecordStore.
     #
-    # To handle either case, we insert it before ActionDispatch::Flash.
+    # To handle either case, we insert it at the end of the middleware stack.
     #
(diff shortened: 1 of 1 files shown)


SamSaffron:
nope, its meant to be at a very particular spot, if its too deep in the stack certain middlewares will mess with it.

tgxworld:
@SamSaffron How about inserting after whatever `Rails.application.config.session_store` is? I think that is way more flexible than assuming the flash middleware is present.

SamSaffron:
That could work, (except for people who disable sessions)

tgxworld:
I think that needs to  be a check that `Rails.application.config.session_store` exists too and raise a message [when it isn't](https://github.com/rails/rails/blob/master/railties/lib/rails/application/default_middleware_stack.rb#L47) since the flash middleware will not be inserted.

tgxworld:
@SamSaffron  I think the follow improves it. Not sure if that was what you had in mind.

```ruby
  initializer "message_bus.configure_init", after: :load_config_initializers do |app|
    MessageBus::MessageHandler.load_handlers("#{Rails.root}/app/message_handlers")

    # We want MessageBus to show up after the session middleware.
    if session_store = app.config.session_store
      app.middleware.insert_after(session_store, MessageBus::Rack::Middleware)
    end

  end
```

Issue #37, opened 2015-06-20
"discourse" is hardcode in redis_channel_name of ReliablePubSub


tony612:
in https://github.com/SamSaffron/message_bus/blob/master/lib/message_bus/redis/reliable_pub_sub.rb#L51

```rb
  def redis_channel_name
    db = @redis_config[:db] || 0
    "discourse_#{db}"
  end
```

Maybe we can have a solution to make this configrable?

SamSaffron:
Sure its a bas prefix, lets just change it to _message_bus

PR #39, opened 2015-06-22
Changes redis_channel_prefix from discourse_ to _message_bus_

elektronaut:
As per #37


SamSaffron:
Yeah this looks good, thanks

PR #41, opened 2015-07-01
Add redis configuration info to readme

st0012:
Hello, recently I am trying to configure my message_bus's redis connection, and I found the api by reading the source code. And I think it would be helpful if everyone can know how to do it by just reading the readme.


SamSaffron:
thanks, more doco always welcome :+1:

PR #42, opened 2015-07-14
Support for json post data

div:
This lets us to make POST with "Content-Type: application/json" and json data. I'm not quite sure what would be the best way to test it cleanly, so I'll be happy to hear your thoughts @SamSaffron.
btw thanks for great gems!


SamSaffron:
interesting, do you have a different client polling?

SamSaffron:
ouch we can not leave this open forever ... @jeremyevans what do you think we should do here? should we also support a json payload? I am not convinced this is needed

jeremyevans:
@SamSaffron With the changes to the rack middleware to support `env['message_bus.channels']`, a user can just use a middleware to process the uploaded JSON data and sets that `env` variable, so I don't see a need for MessageBus to support it natively.

PR #50, opened 2015-09-06
Send flush message to JS when position is ahead

riking:
In certain circumstances, a JS client can get channel positions that are
ahead of the message bus. When this happens, the client will not recieve
any more messages until the page is refreshed. This patch fixes that by:
- (middleware) delivering a message on the __flush channel when the
  condition is detected
- (client) resetting all subscriptions to -1 when the __flush message
  is recieved

Note: The 'ahead' condition happens when the Redis DB is flushed.
Although this is not a common operation, flush-resiliency is a desirable
property of many applications using Redis; this change improves the
flush resiliency of MessageBus from "new messages will be lost for an
indeterminate period of time" to "all old messages are lost, plus those
that happen until the client next polls".

### DIFF
--- a/assets/message-bus.js
+++ b/assets/message-bus.js
@@ -60,16 +60,15 @@ window.MessageBus = (function() {
     var gotData = false;
     if (!messages) return false; // server unexpectedly closed connection
 
-    $.each(messages,function(_,message) {
+    $.each(messages, function(_,message) {
       gotData = true;
-      $.each(callbacks, function(_,callback) {
+      $.each(callbacks, function(_, callback) {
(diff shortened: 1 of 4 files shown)


riking:
Test failure looks spurious.

SamSaffron:
I do want to fix this for 2.0 but I would like a simpler solution if possible.

SamSaffron:
I really want something here but it feels a bit too complicated. 

perhaps if we ever try to subscribe ahead message bus returns the "last id" in status, then client corrects it and if debugging is enabled reports something. Then we don't need a _flush channel and can simply use status

Issue #55, opened 2015-10-23
Clients deleted by timer queue preventing message delivery


robmonie:
I'm seeing an issue where messages sometimes aren't being delivered immediately to clients. After some debugging, I believe what is happening is that the current client instance registered against a client_id in the connection_manager is being removed by timers from previous polls.  This is is particularly evident in our application because we tend to subscribe and unsubscribe to a number of different channels for short periods of time as a user works with the app. We're effectively simulating a request / response type pattern with an event_sourced / cqrs app using promises and short term message_bus subscriptions.

In more detail, the sequence of events that demonstrates this issue is as follows.

1. Subscribe to a channel
  - `middleware` creates a new client and adds it to the `connection_manager` 
  - `connection_manager` registers it against it's `client_id` in the `@clients` attribute of the `connection_manager`
  - `connection_manager` adds the `client_id` to the `@subscriptions` `set`.
  - `middleware` adds callback to the timers queue which calls `remove_client` in `long_polling_interval` seconds. 

2. User subscribes to another channel before the `long_polling_interval` has expired
 - `middleware` creates a new client and subscribes to both channels as above.

3. `long_polling_interval`  for the first subscription expires 
  - `remove_client` method on the `connection_manager` is called. 
  - the client registered against `@clients` in `connection_manager` for the `client_id` is removed. Note that this is actually removing the new client as the first client was replaced by a new client in step 2. This means that any messages published to either of these channels won't be delivered.

4. System publishes a message to the new channel subscribed to in step 2. 
  - No message is delivered to the browser because the client is no longer registered.

5. `long_polling_interval` expires on the client. It re-polls and probably catches up from the backlog. - Note that I haven't completely confirmed this yet.

I have been able to naively stop this problem from happening by checking the `connect_time` of the client being removed against the client currently registered in `@clients`. If they aren't the same i'm not removing the `client` from `@clients` or `client_id` form the `@subscriptions` but only cancelling the `cleanup_timer`. While this seems to help in basic testing, I suspect it may cause other cleanup issues due to remaining subscriptions.

I'll continue to investigate as best I can and update where I can but any thoughts on what i've found would be most appreciated. Especially if i'm off track on what I think i'm seeing.

robmonie:
For reference, this commit has my basic change in it - https://github.com/hooroo/message_bus/commit/7788a52b1d6b75f66e1bb2b8fb2d560a765dce5a

[...]

PR #56, opened 2015-10-27
add spec to demonstrate messages not being delivered to client...

robmonie:
…when subscriptions overlap the cleanup queue timeout.

I hope this test helps to demonstrate #55 . 

Unfortunately the test doesn't quite represent the real world because in a full integration test between the js client and backend, the first polling request would be aborted when the second subscription is created. This second subscription would then poll with both subscriptions. 

AFAIK, there is no way to simulate an aborted http request using rspec. I'm not sure if that invalidates this test or whether there is a way to simulate this to better represent what happens in the real world.

### DIFF
--- a/spec/lib/middleware_spec.rb
+++ b/spec/lib/middleware_spec.rb
@@ -74,6 +74,32 @@ describe MessageBus::Rack::Middleware do
       last_response.headers["FOO"].should == "BAR"
     end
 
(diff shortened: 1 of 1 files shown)


SamSaffron:
I like this example, will have a look at it, sorry for the delay

SamSaffron:
Sorted this out, and added a spec

PR #59, opened 2015-12-12
Simple attempt at bridging message delivery race

benlangfeld:
In the case that a message is published while we're calculating the backlog, we previously missed it. If the backlog turned out to be empty, we would have begun waiting for new messages, but the published message could have been delivered to the connection manager prior to informing it of the client. This would result in the message failing to be delivered until our long-poll timeout.

This adds an in-memory backlog per client which is filled while we calculate the backlog in Redis. The combined (deduplicated) backlog is then considered for delivery, closing the gap. Any messages received between collecting the message buffer and delivering the backlog will become part of the Redis backlog response on the next request.


SamSaffron:
wow, I have been searching for this one for quite a while, great catch, I think it works!

writing a test for this is going to be super tricky. 

Are you using this fix in prd? is it working for you?

SamSaffron:
whats up with all the spec file moving? do we need this?

benlangfeld:
> writing a test for this is going to be super tricky.

Damn right it is, but that's my task for tomorrow :) weekend work is just for the simple stuff ;)

> Are you using this fix in prd? Is it working for you?

Since I only came up with it yesterday, it hasn't made it through our testing process yet. I'll provide notes here on how that goes ASAP.

> what's up with all the spec file moving? Do we need this?

In order for Guard to work correctly and fit with the most common directory layout for ruby gems, yes. See the commit which does it in isolation :)

benlangfeld:
> wow, I have been searching for this one for quite a while, great catch, I think it works!

If there are any other known issues like this, would you mind filing them as issues and assigning to me?

I've not actually proven yet that this is the bug we're encountering in production, and this solution arose from only empirical analysis of the source and stood out as the most obvious flaw. If there are others that those with more experience of MessageBus in production have encountered but not yet solved, I'd like to have a go at squashing those before we encounter them also. In the face of this class of issue I throw scrum out of the window and front-load this kind of thing :)

SamSaffron:
I corrected some long standing concurrency issue recently in https://github.com/SamSaffron/message_bus/commit/b15ba77317ad049856f7dc25a0d397cf43d9bbf3 I do not know of any other issues we have, luckily (or unluckily) the design of message_bus means that it is self healing so we are never hit to hard on these issues. 

Help wise, it would be awesome if we could make CI stop failing erratically, also I would be happy to migrate to minitest cause its a simpler system and a lot of the testing we need is very complex so building on a fat abstraction is not ideal. 

I added some stuff to a "Want to help?" section at the bottom of the readme and any help there would be welcome. 

Appreciate all the help!

PR #61, opened 2016-01-06
Fix concurrency benchmark example

felixbuenemann:
This changes the benchmark to use [wrk](https://github.com/wg/wrk) instead of apache bench for testing concurrency. This is required because we need to generate a unique client id or the server will immediately recycle the client's previous connection, making the benchmark useless.

The client id in the url is generate through the wrk `bench.lua` script.

I've also added a sample puma config.

See #6 for discussion.

Example Output:
```
$ wrk -t2 -c100 -d2m --timeout=45s --latency -s bench.lua http://127.0.0.1:3000
Running 2m test @ http://127.0.0.1:3000
  2 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     1.99s   186.08ms   2.02s    99.11%
    Req/Sec    81.44    108.99   466.00     83.24%
  Latency Distribution
     50%    2.00s
     75%    2.01s
     90%    2.01s
     99%    2.01s
  5948 requests in 2.00m, 1.45MB read
  Socket errors: connect 0, read 5, write 0, timeout 0
Requests/sec:     49.55
Transfer/sec:     12.34KB
```

For realistic tests, wrk should be run from multiple clients against the server, because it's easy to exhaust the dynamic port range on a single machine.


SamSaffron:
This is exciting, any realistic results to share?

SamSaffron:
ok at 4 workers and 10k concurrent we get: 

```
sam@ubuntu bench % wrk -c10000 -d30s --timeout=30s --latency -s bench.lua http://127.0.0.1:9292
Running 30s test @ http://127.0.0.1:9292
  2 threads and 10000 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     2.02s   103.74ms   2.44s    96.08%
    Req/Sec     2.31k     1.78k    8.33k    58.89%
  Latency Distribution
     50%    2.01s 
     75%    2.03s 
     90%    2.05s 
     99%    2.22s 
  87273 requests in 30.05s, 15.65MB read
Requests/sec:   2903.95
Transfer/sec:    533.15KB
```

Not bad at all in my opinion, we service 87k requests in 30 seconds best possible outcome would have been 100k requests cause each was polling for 2 seconds

felixbuenemann:
Not really, but shouldn't be to hard to spin up some digital ocean or aws boxes for testing. On my 13" rMBP with the default max polling interval and 2 unicorn workers I can currently go up to roughly 600 concurrent connections.

I've also checked unicorn vs passenger, which shows that passenger cannot handle more that one request at a time per worker:

```
# config.ru changed to default 25s max polling interval
# unicorn default config with a single worker
$ bundle exec unicorn -E production -l 0.0.0.0:3000 -c config/unicorn.conf.rb
$ wrk -c100 -d2m --timeout=45s --latency -s bench.lua http://127.0.0.1:3000
Running 2m test @ http://127.0.0.1:3000
  2 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    21.89s     8.27s   25.10s    87.47%
    Req/Sec   161.06    193.43   485.00     82.35%
  Latency Distribution
     50%   25.00s
     75%   25.02s
     90%   25.04s
     99%   25.09s
  455 requests in 2.00m, 134.87KB read
  Socket errors: connect 0, read 2, write 0, timeout 0
Requests/sec:      3.79
Transfer/sec:      1.12KB

# passenger standalone default config, limited to one worker
$ bundle exec passenger start --max-pool-size 1 -e production
$ wrk -c100 -d2m --timeout=45s --latency -s bench.lua http://127.0.0.1:3000
Running 2m test @ http://127.0.0.1:3000
  2 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    25.03s     0.00us  25.03s   100.00%
    Req/Sec     0.00      0.00     0.00    100.00%
  Latency Distribution
     50%   25.03s
     75%   25.03s
     90%   25.03s
     99%   25.03s
  4 requests in 2.00m, 1.92KB read
  Socket errors: connect 0, read 0, write 0, timeout 3
Requests/sec:      0.03
Transfer/sec:      16.39B
```
As can be seen passenger only handled 4 requests. The 5th was cut off by the 2m time range.

felixbuenemann:
@FooBarWidget Do you think it would be possible for passenger to reach similar concurrency to puma / unicorn when using rack hijacking? The problem seems to be that passenger doesn't schedule any requests to the worker until the long polling connection is closed.

felixbuenemann:
@SamSaffron Which rack server / config did you use for your benchmark?

SamSaffron:
we should profile to see where the bottlenecks are, once you bump up to 20k concurrent in a puma against a single process we struggle a bit and start seeing a 99 percentile of 10 seconds.

SamSaffron:
Quick round of profiling shows:

```
 TOTAL    (pct)     SAMPLES    (pct)     FRAME
       167  (11.8%)         167  (11.8%)     block in MessageBus::Client#write_and_close
       165  (11.6%)         164  (11.5%)     MonitorMixin#mon_exit
       156  (11.0%)         156  (11.0%)     MonitorMixin#mon_enter
       153  (10.8%)         153  (10.8%)     Puma::Server#normalize_env
       307  (21.6%)         140   (9.9%)     MessageBus::Client#write_and_close
       134   (9.4%)         134   (9.4%)     MessageBus::TimerThread::Cancelable#initialize
       168  (11.8%)          81   (5.7%)     Puma::Server#handle_servers
       180  (12.7%)          80   (5.6%)     Puma::Client#eagerly_finish
       101   (7.1%)          80   (5.6%)     Puma::Client#try_to_finish
        64   (4.5%)          63   (4.4%)     block in Puma::ThreadPool#<<
       185  (13.0%)          31   (2.2%)     MessageBus::TimerThread#do_work
       677  (47.6%)          25   (1.8%)     MessageBus::Rack::Middleware#call
        21   (1.5%)          21   (1.5%)     Puma::Client#setup_body
        13   (0.9%)          13   (0.9%)     Puma::Client#initialize
        18   (1.3%)          10   (0.7%)     Rack::Utils#parse_nested_query
         6   (0.4%)           6   (0.4%)     Rack::Utils::KeySpaceConstrainedParams#initialize
        87   (6.1%)           5   (0.4%)     block in Puma::Server#handle_servers
```

I wonder if we can async write and async close ....

felixbuenemann:
Btw. I think your wrk version might be outdated, it doesn't show the "Socket errors:" line which is interesting (or maybe it's hidden if there are no errors). I'm using wrk 4.0.

SamSaffron:
I just built it from github, maybe there is a branch or something ...

felixbuenemann:
Nah just checked, it's hidden if there are no errors. And I should really test on Linux, Mac OS X has poor network concurrency, even with sysctl tuning.

SamSaffron:
yeah its super weird that mac is the worstest when it comes to concurrent io you would think with kqueue and all the underlying bsd goodness it would be the bestest.

felixbuenemann:
:-) Besides the fact that eg. thin doesn't use kqueue although eventmachine supports it. But most of the defaults are tuned for a desktop os. There's a lot of stuff that can be tuned through sysctls though.

felixbuenemann:
Btw. on puma I sometimes get this exception during benchmarking:
```
2016-01-06 04:15:30 +0100: Rack app error: #<NoMethodError: undefined method `write' for nil:NilClass>
/home/felix/.rvm/gems/ruby-2.2.4/bundler/gems/message_bus-b7d13ea391e0/lib/message_bus/client.rb:197:in `write_chunk'
/home/felix/.rvm/gems/ruby-2.2.4/bundler/gems/message_bus-b7d13ea391e0/lib/message_bus/client.rb:48:in `ensure_first_chunk_sent'
/home/felix/.rvm/gems/ruby-2.2.4/bundler/gems/message_bus-b7d13ea391e0/lib/message_bus/rack/middleware.rb:136:in `block in call'
/home/felix/.rvm/gems/ruby-2.2.4/bundler/gems/message_bus-b7d13ea391e0/lib/message_bus/client.rb:20:in `block in synchronize'
/home/felix/.rvm/gems/ruby-2.2.4/bundler/gems/message_bus-b7d13ea391e0/lib/message_bus/client.rb:20:in `synchronize'
/home/felix/.rvm/gems/ruby-2.2.4/bundler/gems/message_bus-b7d13ea391e0/lib/message_bus/client.rb:20:in `synchronize'
/home/felix/.rvm/gems/ruby-2.2.4/bundler/gems/message_bus-b7d13ea391e0/lib/message_bus/rack/middleware.rb:133:in `call'
```
Maybe this happens when the client closes the socket prematurely (`@io` is nil)?

felixbuenemann:
We need to set `PhusionPassenger.advertised_concurrency_level = 0` when using passenger, see:

https://github.com/phusion/passenger/issues/1187#issuecomment-44814845 and phusion/passenger#1195.

I'll make a PR to update the README and benchmark.

felixbuenemann:
Hmm, this should already be auto enabled by message_bus, but maybe it needs to be done before booting the rack app. Because the benchmark I posted above improves a lot if I enable it in the config.ru of the benchmark:

```
$ bundle exec passenger start --max-pool-size 1 -e production
$ wrk -c25 -d2m --timeout=45s --latency -s bench.lua http://127.0.0.1:3000
Running 2m test @ http://127.0.0.1:3000
  2 threads and 25 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    20.52s     9.63s   25.04s    82.05%
    Req/Sec    27.67     44.32   120.00     75.00%
  Latency Distribution
     50%   25.00s
     75%   25.01s
     90%   25.02s
     99%   25.04s
  117 requests in 2.00m, 54.33KB read
Requests/sec:      0.97
Transfer/sec:     463.55B
```

evanphx:
I looked at why `normalize_env` is taking so much time. Turns out it's the call to `IPSocket#peeraddr`, specifically when it does the `getpeername` syscall. Sadly there is no way to optimize that away except for just deferring it, but since it's almost always accessed during a request (for instance, to log the client address) deferring only makes things slower.

evanphx:
I could add an option to puma that disables using `#peeraddr` entirely to speed up apps that don't need need it.

SamSaffron:
@evanphx  this is actually a fairly interesting and generally applicable problem that could possibly be optimised. 

vast majority of deployments will have NGINX or HAProxy -> puma, so the peer address really is fairly uninteresting as all you really care about is the x-forwarded-for address, I wonder if perhaps a setting would be applicable here, clearly not a default on thing but makes sense that in some cases you just don't care who the peer is.

felixbuenemann:
Also keep in mind, that normally you don't get 10k simultaneous connections at once, but rather connections building up over time. One interesting thing is to have some kind of randomization or backoff in the client to avoid trashing a server with connections when a restart occurs.

Maybe we can look for or write a benchmark that gradually ramps up connections.

tinco:
Hi, just pitching in to say that the advertised_concurrency_level has to be set at the rack bootup, as it is read only once by passenger as it prepares for spawning workers. I just ran the benchmarks for puma, unicorn and passenger and they all perform within 15% of each other in the -c100 benchmark. I just added the advertised concurrency level setting in the config.ru file right after the `if defined?(PhusionPassenger)` line.

felixbuenemann:
@tinco Thanks for the confirmation, I'll make a PR with the required changes. One thing I noticed with `advertised_concurrency_level = 0` is that passenger never spun up more than a single worker, I haven't yet checked if this interferes with normal app requests if running the middleware inside a rails app.

OnixGH:
@felixbuenemann yeah, when you set unlimited concurrency, autoscaling will not work (since it is based on max concurrency). One way around it is to set it to a concrete number, and once that level of concurrency is reached, a new instance will be autospawned.

If you apply it to an app with normal requests that will also work, but it's not recommended since passenger will pass the requests straigth to app processes and they will do the queueing, meaning that load balancing can become sub-optimal.

evanphx:
How about a config setting called 'remote_addr' that can having the values ':socket' (the default), ':forwarded' to pick up the X-Forwarded value, or ':localhost' to just set it to "127.0.0.1". Thoughts?

felixbuenemann:
@OnixGH That's problematic, because if the message_bus middleware runs inside the rails app, only certain routes handled by the middleware have high concurrency, while all other routes have normal concurrency. So essentially to work around this, one has to run a separate app that only serves the message_bus middleware.

felixbuenemann:
@evanphx Generally there are two types of headers, 'X-Real-IP' that contains the client IP and 'X-Forwarded-For' which contains the client ip and chain of intermediate proxies.

Example:
```
X-Real-IP: 54.83.132.159
X-Forwarded-For: 54.83.132.159, 199.27.72.25, 50.19.19.94
X-Forwarded-Port: 80
X-Forwarded-Proto: http
```

So as settings we could used `:socket`, `:x_forwarded_for`, `:x_real_ip` and maybe if a string is used it must be an ip v4 or v6 address (`'127.0.0.1'`, `'::1'`).

evanphx:
https://github.com/puma/puma/commit/d5914573c90fe6f8b9422f397e28c0358ce4aa3b

If you want, switch to puma out of git and try your benchmark again with `set_remote_address` set to something other than `:socket`. If this seems to help, i'll go ahead and release it today!

OnixGH:
@felixbuenemann actually, you don't have to run a separate app, because you apply the concurrency option to a location block that is also an app group. See the [passenger_force_max_concurrent_requests_per_process](https://www.phusionpassenger.com/library/config/nginx/reference/#passenger_force_max_concurrent_requests_per_process) we introduced in 5.0.22.

OnixGH:
An example from @FoobarWidget:
```
server {
   listen 80;
   server_name foo.com;
   root /webapps/foo/public;
   passenger_enabled on;

   # Use default concurrency for the app. But for the endpoint /special_websocket_endpoint,
   # force a different concurrency.
   location /special_websocket_endpoint {
       passenger_app_group_name foo_websocket;
       passenger_force_max_concurrent_requests_per_process 0;
   }
}
```

felixbuenemann:
@OnixGH I just tried that example on 5.0.23 and it crashes the passenger core as soon as a second concurrent request hits the location blocks app group… I'll make a bug report in the passenger repo.

FooBarWidget:
The Passenger crash bug has been fixed and the fix will be included in 5.0.24.

felixbuenemann:
Yes, I'm still waiting on 5.0.24 to be released before I make a PR with the suggested configuration changes.

Issue #57, opened 2015-11-06
Can not start puma when set MessageBus.after_fork on boot work


embbnux:
when I deploy my rails app in production with puma, puma socket cannot start if I add:
```ruby
on_worker_boot do
    MessageBus.after_fork
end
```
anything wrong? Thank you
log:
```ruby
[11443] Puma starting in cluster mode...
[11443] * Version 2.14.0 (ruby 2.2.3-p173), codename: Fuchsia Friday
[11443] * Min threads: 4, max threads: 16
[11443] * Environment: production
[11443] * Process workers: 1
[11443] * Phased restart available
[11443] * Listening on unix:///tmp/app.sock
[11443] * Daemonizing...
```
Ang Nginx cannot connect to unix:///tmp/app.sock

SamSaffron:
Is anything in the logs? do you have redis installed?

embbnux:
thank you, redis installed.puma do not log everything after [11443] * Daemonizing...,
but nginx cannot connect to puma socket.
If I remove 'MessageBus.after_fork' and restart, it works well

SamSaffron:
Sorry for the late reply, what happens if you run it in not forking mode, does it work correctly?

felixbuenemann:
@embbnux Add `require "message_bus"` just before the `on_worker_boot` block.

SamSaffron:
yeah this has to be a user error, we work fine in a forked environment.

SamSaffron:
we should probably update the readme though.

embbnux:
thank you ! it work well

jstdoit:
but, after I added the `require message_bus` in unicorn production mode, in browser, it got `404`

PR #62, opened 2016-01-06
[ci skip] Fix README examples for puma and unicorn

felixbuenemann:
This makes sure that `::MessageBus` is defined. See #57.

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,6 +214,7 @@ end
 #### Puma
 ```ruby
 # path/to/your/config/puma.rb
+require 'message_bus'
 on_worker_boot do
   MessageBus.after_fork
 end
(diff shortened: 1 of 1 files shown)


SamSaffron:
Thanks!

tony612:
I think there's something wrong with the require for unicorn. When I use `require`, I got this error
```
undefined method `[]' for nil:NilClass backtrace: ["/home/deployer/apps/playdock_production/shared/bundle/ruby/2.2.0/gems/message_bus-1.1.1/lib/message_bus.rb:465:in `block (2 levels) in global_subscribe_thread'"
...
```

Then even if I add `@subscriptions ||= {}` before these lines, it still doesn't work. It seems the MessageBus wasn't loaded into my rails app, I always got 404 for message bus requests

felixbuenemann:
The require in the unicorn/puma configs only ensures that the after fork hooks work. They have nothing to do with loading the middleware inside your app, so it's a different issue.

PR #67, opened 2016-01-25
Explicitly say that message-bus can be required

Vaporizd:
So people know if you can include message-bus via asset pipeline.


SamSaffron:
Thanks!

PR #68, opened 2016-01-28
Link to the Ruby implementation of the client.

lowjoel:


SamSaffron:
Thanks!

SamSaffron:
Is the client handling the chunked encoding stuff?

lowjoel:
Yup, chunked encoding works.

I'm assuming chunked vs non-chunked encoding is that the non-chunked encoding simply returns one message (or array) then immediately drops the connection?

In that case i'll just immediately create another long polling connection.

However, there isn't any fallback logic that the JS client has (like long polling + chunked, then long polling, then polling) because I imagine that whoever's using the gem would be deploying this on a server and would know his network infra well?

SamSaffron:
really depends on the use case, if you are using this internally no need to worry about fallbacks, if this becomes bundled in some sort of tool then the fallbacks do become important.

SamSaffron:
I would not rush to implement everything though, wait to see what people ask for.

lowjoel:
Precisely :)

PR #71, opened 2016-02-12
Port the test suite to MiniTest

jeremyevans:
This actually uses minitest/spec.  Hopefully that's fine.

This replaces rspec with minitest in the Gemfile, and removes
the redis, rack, rbtrace, guard-rspec, and rb-notify gems from
the Gemfile.  redis and rack in the gemspec and do not need to
be specified twice, guard-rspec seems to be related to rspec,
and rbtrace and rb-inotify don't seem to be needed to run the
tests.

Most of this was done mechanically using the same script I
used for converting Sequel's and all of my other libraries'
specs from rspec to minitest.

I used require_relative in the specs to get the spec_helper.
This should be fine as messagebus doesn't appear to support
ruby 1.8.

### DIFF
--- a/spec/lib/message_bus/timer_thread_spec.rb
+++ b/spec/lib/message_bus/timer_thread_spec.rb
@@ -1,4 +1,4 @@
-require 'spec_helper'
+require_relative '../../spec_helper'
 require 'message_bus/timer_thread'
 
 describe MessageBus::TimerThread do
@@ -13,7 +13,7 @@ describe MessageBus::TimerThread do
   it "allows you to queue every jobs" do
     i = 0
     m = Mutex.new
-    every = @timer.every(0.001){m.synchronize{i += 1}}
+    every = @timer.every(0.001){m.synchronize{i += 1 if i < 3}}
     # allow lots of time, cause in test mode stuff can be slow
     wait_for(1000) do
       m.synchronize do
@@ -22,14 +22,14 @@ describe MessageBus::TimerThread do
       end
     end
     sleep 0.002
-    i.should == 3
+    i.must_equal 3
   end
 
   it "allows you to cancel timers" do
     success = true
     @timer.queue(0.005){success=false}.cancel
     sleep(0.006)
-    success.should == true
+    success.must_equal true
   end
 
   it "queues jobs in the correct order" do
@@ -45,7 +45,7 @@ describe MessageBus::TimerThread do
       4 == results.length
     }
 
-    results.should == [0,1,2,3]
+    results.must_equal [0,1,2,3]
   end
 
   it "should call the error callback if something goes wrong" do
@@ -67,7 +67,7 @@ describe MessageBus::TimerThread do
       error
     end
 
-    error.class.should == NameError
+    error.class.must_equal NameError
   end
 
 end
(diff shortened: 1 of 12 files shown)


spec/lib/message_bus/timer_thread_spec.rb
```diff
@@ -13,7 +13,7 @@
   it "allows you to queue every jobs" do
     i = 0
     m = Mutex.new
-    every = @timer.every(0.001){m.synchronize{i += 1}}
+    every = @timer.every(0.001){m.synchronize{i += 1 if i < 3}}
```
SamSaffron: good catch

SamSaffron:
Thanks @jeremyevans  much appreciated, I still want to get guard going but this is not blocking me merging this

Are you using MessageBus? Super happy for any general feedback here.

jeremyevans:
I'm not using MessageBus in production yet, but from my experience testing it, it's much simpler and more reliable than websockets.  I'm considering adding a plugin for Roda to use MessageBus (it already has one for websockets).

One thing I see may be an issue is there doesn't appear to be per-channel security settings.  You can limit messages by user_id and group_id, but my preferred way to handle security would be to not even accept requests for channels that the user doesn't have access to.  Ideally, access control would be bit into the web application itself.  However, MessageBus's rack middleware approach doesn't really work with that.  I'd like to have an API in Roda like so (for a multi-user, multi-room chat server):

```ruby
  route do |r|
    r.on "room/:id" do |room_id|
      room = Room[room_id.to_i]
      r.on "message_bus" do
        if logged_in_user.has_access_to?(room)
          r.run MessageBusApp
        end
      end
    end
  end
```

I think that's something I can do with MessageBus's current API, it's just going to take a little more work.

I've already started work on a PostgreSQL backend for MessageBus, which works by exposing the Redis API subset that MessageBus uses.  I'll submit a pull request for that if I'm able to get it working.

One of the other things under "Want to help?" in the README is an in memory storage backend.  I haven't tried it, but isn't it possible that FakeRedis (https://github.com/guilleiguaran/fakeredis) could take care of that without significant changes to MessageBus?

SamSaffron:
I was thinking about the memory only implementation as a clear minimal example of what the interface for reliable pub sub can be (which can also be used in a standalone puma for people who are happy with single process), FakeRedis does not feel like a great fit for me, I much prefer to test using real redis. 

PG adapter sounds absolutely awesome, looking forward to seeing it! 

Regarding the endpoint design, at the moment message_bus like to multiplex multiple channels over a single HTTP endpoint, this heavily reduces the amount of HTTP calls you need to make to keep track of multiple subscriptions. For your chat example, it works fine if all the channels involved live in the chat room, but as soon as you add more channels to the mix (like general notifications and app is out of data channel and so on) HTTP calls you need to make can explode. 

That said, we can run multiple MessageBuses per app, and clean up the middleware to be reusable across multiple endpoints, stuff can be broken down to that level without adding much code.

jeremyevans:
Note that I'm not advocating switching to FakeRedis for MessageBus's tests.  I was thinking it would be useful for simple single-process deployments.  I haven't tried, but hopefully using FakeRedis will still pass all tests except the multiprocess ones.  Assuming it does, I think it would be simpler to use that compared to building your own in-memory implementation from scratch.

I can certainly see how using a single HTTP endpoint can be better for performance if your use case is subscriptions to multiple channels. Hopefully adding support for per-channel security will be possible without affecting the existing support for that.

PR #72, opened 2016-02-23
Add PostgreSQL backend using LISTEN/NOTIFY

jeremyevans:
This supports PostgreSQL as a backend.  Originally, I attempted to
just provide a replacement Redis object that duck-typed and worked
on PostgreSQL, but due to differences between PostgreSQL and Redis
I was not able to get it to reliably pass the multi_process_spec.

This adds a lib/message_bus/postgres/reliable_pub_sub.rb file
with the PostgreSQL support.  To use the support, you currently
need to require that file manually, then pass a
:pub_sub_class=>MessageBus::Postgres::ReliablePubSub option in
your redis_config.  Connection options to PostgreSQL go in a
:pg_config subhash.

One significant difference between this PostgreSQL backend and the
Redis one is the dropping of channel-specific backlog ids. The
PostgreSQL backend only stores a single backlog id, not separate
global and channel backlog ids.  I don't think the lack of a
channel specific backlog id should be a problem. As long as the
channel backlog id is monotonically increasing, no messages
should be missed.

This does make some minor changes to the specs to make them support
both Redis and PostgreSQL.  Most of the changes are simple and
self explanatory.  I did add a ensure exit!(0) to the specs that
fork.  You can't call a normal exit in these specs, as the
garbage collector messes with the database connection sockets,
you need to call exit!(0) to skip that.  I added the ensure because
if there is an exception, it will be caught by minitest, but the
specs will keep running in both the child and parent processes.

Currently, the PostgreSQL backend does not automatically create the
table for you on first use, you need to do that manually by calling
MessageBus::Postgres::ReliablePubSub.setup!, which also clears the
table if it already exists.

This PostgreSQL backend reliably passes all specs for me, and works
for me in a simple application.  However, I didn't do any testing
with it at scale.


lib/message_bus.rb
```diff
@@ -182,7 +181,12 @@ def reliable_pub_sub=(pub_sub)
   def reliable_pub_sub
     @mutex.synchronize do
       return nil if @destroyed
-      @reliable_pub_sub ||= MessageBus::Redis::ReliablePubSub.new redis_config
+      @reliable_pub_sub ||= if redis_config[:pub_sub_class]
+        redis_config[:pub_sub_class].new redis_config
```
SamSaffron: redis_config seems a bit odd here especially if sent to a non redis class...

lib/message_bus/postgres/reliable_pub_sub.rb
```diff
@@ -0,0 +1,445 @@
...
+    @pid = Process.pid
+    reconnect
+  end
+
+  def sync
+    @mutex.synchronize{yield}
+  end
+
+  def hold
+    if Process.pid != @pid
```
SamSaffron: nice that you added fork detection!

lib/message_bus.rb
jeremyevans: Agreed, but changing that requires changing the MessageBus API.

SamSaffron:
This is fantastic! 

A few points 

- I would like to get rid of the redis legacy here, the api should be message buses api not redises... 

- we need to update the readme to explain usage

- we should strip the runtime dependency on redis

SamSaffron:
Yep fine to make this a 2.0 release and break APIs if needed, this is better for the project.

SamSaffron:
technically we are in 2.0 beta at the moment, so breaking APIs is ok ...

jeremyevans:
That sounds good.  Which parts do you want me to take care of?  There's still a lot of redis-specific naming in the PostgreSQL backend, but that should be fairly simple for me to remove.   I didn't do that yet to make diffing between the redis backend and postgres backend easier.

In terms of the non-PostgreSQL specific code, such as changing `redis_config` to `config`, is that something you would like to take care of? 

In terms of README changes explaining how to use the PostgreSQL support, I think the way it works now is a little suboptimal.  But since you are open to API change, I think something like:

```ruby
MessageBus.config(:redis) # default
BUS = MessageBus::Instance.new
BUS.config(:postgres, :user=>'...', :dbname=>'...') # connection options
```

would be better than the current API.  `config` here would set the default backend and connection options for the bus.  This would require `message_bus/backends/#{backend}`, allowing support for external backends.  If `config` hasn't been called before `reliable_pub_sub` is needed, then redis will be assumed to not needlessly break backwards compatibility.

SamSaffron:
Ideally if you could take care of the kaboodle, I am really struggling for
time over the next 4 weeks as I am on the move...

Totally open to that API change sounds good to me, but I would extend it
somewhat...

group_id_lookup, user_id_lookup, site_id_lookup are all configuration
settings and should be fed in via the config object.

perhaps

MessageBus.config(
  backend: {store: :pg, setting...}
  user_id_lookup: proc{|env| ... }
)

etc...

jeremyevans:
OK.  I'll try to work on this starting tomorrow.

SamSaffron:
Awesome !

jeremyevans:
I think I have an idea why this is occasionally failing on Travis.  I'm having a hard time reproducing it locally, with 99/100 times it passing, probably due to faster hardware.  Tomorrow I'm going to try refactoring the PostgreSQL backend to always use a new connection for the LISTEN, instead of one from the pool.  I think the unsubscribe logic can also be simplified, and the fork detection logic in the connection pool and during LISTEN can probably be improved.  It may also be that GC is being run in the child process and causing issues in the parent due to sharing the socket (see https://github.com/ged/ruby-pg/blob/master/ext/pg_connection.c#L167).  The only way to fix that currently is to always keep a reference to any inherited sockets in child processes, so that they never get GCed.  Another way to fix that would be to change ruby-pg to only call `PQfinish` on the connection if the current pid is the pid that created the connection.  I'll see if the ruby-pg maintainers would be amenable to that.

jeremyevans:
@SamSaffron  I think this is ready to be merged.  No errors when running locally with MULTI_PROCESS_TIMES=500.  I'd like to get your review and OK before I merge this.

jeremyevans:
After removing the sleep call from the multi_process_spec and running the specs again, I had one failure after a few hundred runs, with only 99/100 messages received.  This turned out to be caused by another race issue in the postgres backend due to INSERT vs. COMMIT timing. I've tested this with MULTI_PROCESS_TIMES=2000 without failures.

SamSaffron:
Awesome lets merge this in. 

Will do a new beta on Monday, and amend Discourse to use the new APIs

jeremyevans:
Great! I started work on an in-memory backend today.  I originally tried fakeredis with the existing redis backend, but the latest release is 18 months old and doesn't support publish/subscribe, and the master branch didn't work with the current redis backend.  I hope to have the in-memory backend ready by early next week, it's already at only 6-7 spec failures.

I think the APIs are stable, so releasing a beta is fine, but I'd like the 2.0 final release to include the in-memory backend, assuming I can get it working in time.  What's your ETA for a 2.0 final release?

SamSaffron:
ETA wise, I would like to run beta in production for at least 2 weeks
before stamping out 2.0 final, so let's aim for 2-3 weeks from now, does
that work for you?

jeremyevans:
Sure, that's fine.  I should have the in memory backend finished by mid next week, it's a lot simpler than the postgres backend.

jeremyevans:
I've been torture testing the postgres backend since Friday afternoon, with MULTI_PROCESS_TIMES=100000.  Results:

```
Finished in 137932.223651s, 0.7255 runs/s, 1.4508 assertions/s.

100073 runs, 200118 assertions, 0 failures, 0 errors, 1 skips
```

I'm confident there are no race conditions left.

SamSaffron:
Wow, super duper awesome.

SamSaffron:
I did a release, have a look through my tiny commits, I changed the default pg user for test and disabled verbose logging in redis ... upgrade was totally painless, so not on breaking change really is not needed.

https://github.com/discourse/discourse/commit/5dd746d61f7298b12830b2a1d0f82736c2d52fdc 

Can you review changelog for your changes and clean up to taste? 

:+1:  on all your work here it is Awesome

jeremyevans:
Changes look good.  I was planning on renaming the `MESSAGE_BUS_REDIS_CONFIG`, but looks like you beat me to it.

One possible feature I was thinking of adding to the postgres backend is an option for only clearing the backlog every x number of messages (e.g. every 100 or 1000), instead of on every message.  I haven't benchmarked it yet, but I'm guessing that would significantly improve publish speed. If you don't need exact backlog clearing it's probably a good tradeoff. Thoughts on that?

SamSaffron:
yes definitely, a low / high watermark is a much better policy there, got
to be careful with deadlocks when deleting batches though (and possibly
delete in a transaction so its faster)

SamSaffron:
also just thinking out loud about an edge cases we may hit, what happens
when message number 2147483647 is hit?

jeremyevans:
In the postgres backend, the deletion is on rows where `id < $oldest_id_to_keep`.  So this wouldn't change the queries at all, it just wouldn't run the deletion query on most publish calls.

And I used a 64-bit counter, not a 32-bit one, so we can basically ignore wrap around. :)

SamSaffron:
Nice, looks good to me

We should try to bench performance of redis vs pg for release cause that
would also help guide people to picking the right backend for the job

PR #73, opened 2016-02-24
Support passing channels to Rack middleware via env['message_bus.channels']

jeremyevans:
This changes MessageBus::Rack::Middleware to recognize the
'message_bus.channels' env entry as the channels to use for the
request, in preference to creating a new Rack::Request object
and using Rack::Request#POST.

This makes it easier for earlier middleware or request handler to
handle access controls for channels.  Currently, to do so would
require messing with rack.input, which is kind of a pain.

To support setting the client.seq value, that can also be specified
as the 'message_bus.seq' env entry.

There is no change to backwards compatibility if the
'message_bus.channels' env entry is not present.

For an example of how I plan to use this with Roda:
http://pastie.org/pastes/10736290/text

### DIFF
--- a/lib/message_bus/rack/middleware.rb
+++ b/lib/message_bus/rack/middleware.rb
@@ -84,12 +84,21 @@ class MessageBus::Rack::Middleware
     client = MessageBus::Client.new(message_bus: @bus, client_id: client_id,
                                     user_id: user_id, site_id: site_id, group_ids: group_ids)
 
-    request = Rack::Request.new(env)
-    request.POST.each do |k,v|
-      if k == "__seq".freeze
-        client.seq = v.to_i
-      else
+    if channels = env['message_bus.channels']
+      if seq = env['message_bus.seq']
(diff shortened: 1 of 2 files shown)


SamSaffron:
Sure, looks good!

SamSaffron:
cool, also I have given you commit here in case you have any small fixes that you need to apply etc. :)

benlangfeld:
Would most apps not prefer to define a callback like the user/group lookup to do access control? Is there another use case for this feature?

SamSaffron:
Yes, however this allows us flexibility of doing auth in say NGINX LUA and then handling message bus in Ruby which is pretty enormous imo.

benlangfeld:
@SamSaffron How are lower-case values in the Rack::Env set from outside the Ruby process?

SamSaffron:
True, we should probably amend it so the feature is opt-in and then use a
proper HTTP header, it would keep both use cases working fine

PR #74, opened 2016-02-29
Make all specs run with same backend, fix PostgreSQL fork handling

jeremyevans:
redis_config= was used in the specs, which sets the backend to redis.  This resulted in some of the specs being run on redis even when using `spec_postgres`.  Fix that, then fix a couple issues exposed when all specs run with the postgres backend.  One was a spec issue, one was an issue in the postgres backend related to fork handling.

### DIFF
--- a/lib/message_bus/backends/postgres.rb
+++ b/lib/message_bus/backends/postgres.rb
@@ -103,25 +103,19 @@ class MessageBus::Postgres::Client
     sync{@listening_on[channel] = obj}
     listener = Listener.new
     yield listener
-    pid = Process.pid
     
     conn = raw_pg_connection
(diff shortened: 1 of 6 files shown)


jeremyevans:
There's a single spec failure on ruby 2.3.0 only.  I can't reproduce it even with 1000 iterations.  It appears the postgres backend is providing an empty value entry, which shouldn't happen.  We could probably catch this closer to the source by adding a check constraint to the column.  I think the error might be because the bus doesn't call reset! before the spec, so I'll do that as well.

jeremyevans:
I think my changes fixed the issue, so I'll merge this now.

PR #75, opened 2016-02-29
Add memory backend

jeremyevans:
The memory backend uses the same specs as the postgres backend,
except it doesn't run the fork or multi_process specs, since
those wouldn't work.

Only change to the lib code is returning instead of raising on
a destoryed bus.  The memory backend is fast enough that it
exposes a race condition in the rack middleware spec causing
warnings but no failures.  I think it's simpler to just
return in this case, as if the bus has been destroyed, you
don't want to keep subscribing.


SamSaffron:
awesome!

PR #76, opened 2016-02-29
Support :clear_every option on postgres and memory backends

jeremyevans:
This option will only clear the backlog every x requests,
which can increase performance on the postgres backend
by skipping 2 queries most calls to publish.

This also removes unused options from all backends.


SamSaffron:
nice one, and nice cleanup!

PR #77, opened 2016-03-02
FEATURE: Allow Redis config to be updated for `MessageBus::Redis::Rel…

tgxworld:
…iablePubSub`.


jeremyevans:
I don't think the travis failure is related to this commit.  But you should at least add a spec for this.

SamSaffron:
Agree we need a spec, also we should flow this in from the top, feels odd
to give this feature only to one backend.

Another option here is to do nothing and simply throw away the bus instance
when reconfigure is needed, feels more correct to me

PR #80, opened 2016-03-04
Update README.md

michalhorcic:
Adding information about solving slow application under passenger with message_bus running


SamSaffron:
thanks, will clean up the docs based on the example

Issue #64, opened 2016-01-14
Where do user_ids and group_ids get set?


jemminger:
(is there some better place to ask for help?  Stackoverflow only has 12 followers of the "discourse" tag)

The readme shows this example:

```ruby
# messages can be targetted at particular users or groups
MessageBus.publish "/channel", "hello", user_ids: [1,2,3], group_ids: [4,5,6]
```

However it does not make it clear what specifically a "user_id" is or where it is set.  In my use case, I want to broadcast a message to a specific user.  Is the user_id the user's database ID, or something else?  

I suspect it's just some arbitrary identifier that I can just set into the current session and retrieve via 

```ruby
  MessageBus.user_id_lookup do |env|
    # return the user id here
  end
```

but I cannot seem to find the right place to put the `user_id_lookup` call where the rails session actually exists yet.  All I get for `env["rack.session"]` is `#<ActionDispatch::Request::Session:0x7fb788510dc8 not yet loaded>`

SamSaffron:
run `rake middleware` to see where MessageBus is in the stack, you may want
to move it around to a more appropriate place if its too early in the
stack.

The most efficient spot for it is super early on, but you can always put it
way further down.

Once you sort this can you do a PR to improve the readme ?

JAndritsch:
I managed to configure it via an initializer (config/initializers/message_bus.rb). My code looks something like this:

```
MessageBus.user_id_lookup do |env|
  # use the incoming env object to determine what this user's ID should be
  # ex:
    request = Rack::Request.new(env)
    request.cookies["user_id"]
end
```
In that example (which is definitely not the most secure way to do this), you parse the incoming ENV object into a Rack::Request, which then lets you ask for the request cookies, params, or even headers. 
This example expects that each user who connects with MessageBus will have their unique "user_id" cookie set at the time of the request. 

Again, not the most secure code since cookies can be altered but the example was meant to illustrate how this is done.

The same thing can be done with group identification. Here's a basic example:

```
MessageBus.group_ids_lookup do |env|
  # use the incoming env object to determine what groups this user should have
  # ex:
    request = Rack::Request.new(env)
    user_id = request.cookies["user_id"]
    user = User.find(user_id)
    user.roles # => [ "admin", "super user", "something else" ]
end
```
Does that help?

SamSaffron:
> Does that help? 

yes it does :) we should get something to help cover this in the readme.

SamSaffron:
[...]

PR #82, opened 2016-03-06
Improve documentation on id lookups

JAndritsch:
Hopefully this helps clear up the usage of `user_id_lookup` and `group_ids_lookup`. 

Reference: #64

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,9 +61,9 @@ MessageBus.publish "/channel", "hello", client_ids: ["XXX","YYY"]
 # MessageBus determines the user ids and groups based on env. How that data is calculated is entirely up to you.
 
 MessageBus.configure(user_id_lookup: proc do |env|
-  # Supply the method body for extracting and returning a unique user id when given the env.
+  # Supply the method body for extracting and returning a unique user id.
 
-  # Example: Retruning a user_id cookie from the request.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -58,15 +58,29 @@ MessageBus.publish "/channel", "hello", user_ids: [1,2,3], group_ids: [4,5,6]
...
-# message bus determines the user ids and groups based on env
+# MessageBus determines the user ids and groups based on env. How that data is calculated is entirely up to you.
 
 MessageBus.configure(user_id_lookup: proc do |env|
-  # return the user id here
+  # Supply the method body for extracting and returning a unique user id.
+
+  # Example: Returning a user_id cookie from the request.
+  request = Rack::Request.new(env)
+  request.cookies["user_id"]
```
SamSaffron: this is very risky as a sample, storing a user_id in a cookie would be real bad... instead perhaps

u = lookup_user(env)
u && u.id

ffabreti: Can we query database and retrieve post params from inside user_id_lookup ? Something like:
```
req = Rack::Request.new(env)
if  SocialUser.validate_token(req.params["user_id"], req.params["user_token"])
   req.params["user_id"]
else
   nil
end
```

SamSaffron:
closing for now, let me know if you want to give it another shot.

PR #84, opened 2016-03-07
FIX: MessageBus stucked in infinite loop.

tgxworld:


lib/message_bus/backends/redis.rb
```diff
@@ -167,13 +177,6 @@ def ensure_backlog_flushed
 
         @in_memory_backlog.delete_at(0) unless try_again
       end
-
-      if try_again
-        sleep 0.005
-        # in case we are not connected to the correct server
```
SamSaffron: why dod we get rid of the sleep here, don't want us hitting redis really hard during readonly

lib/message_bus/backends/redis.rb
```diff
@@ -147,11 +147,21 @@ def publish(channel, data, queue_in_memory=true)
...
 
   def ensure_backlog_flushed
-    while true
+    flushed = false
+
+    while !flushed
       try_again = false
 
+      if is_readonly?
+        sleep 1
```
SamSaffron: ahh it is here .... I see.

SamSaffron:
@jeremyevans this is a bit of fancy we probably want to add in to the pg adapter as well, not something we need for 2.0 release but possibly at some point

jeremyevans:
@SamSaffron In memory backlog support is not present in the postgres or memory backends, only in the redis backend, so this wouldn't apply to the postgres backend.

SamSaffron:
Sorry, I should have expanded.

PG can also be in a readonly status during failover, so it *may* make sense
to add support for an in-memory backlog, not something that we need right
now, but a possibility.

jeremyevans:
I get the feeling that an in memory backlog is less important for PostgreSQL than for Redis.  In general, if PostgreSQL is being used, it's probably because it's also the main database for the site and the user wants to reduce the number of dependencies.  If the PostgreSQL connection is down, the whole site is probably down.  This isn't likely to be true for Redis, because few people would use Redis as the main database for a site.

That said, I'm not against supporting it.  The reason it isn't supported in the postgres backend:

1. No tests for the feature
2. The redis backend only uses it when there is a Redis specific exception which I'm not sure there is a PostgreSQL equivalent.

PR #86, opened 2016-03-10
Replace jQuery.ajax with XMLHttpRequest requests and some Jasmine client tests.

nathanstitt:
This is a work in progress that I'm submitting early in order to get feedback on if it's a good idea and might be accepted, and in order to garner suggestions on improvements.

Some background:  I've been using message-bus for some time on a React based project of mine that doesn't use jQuery.  Since I didn't have it available, I hacked together [an implementation](https://github.com/argosity/lanes/blob/master/client/lanes/vendor/message-bus.coffee) that used the `xhr` js library (what the project uses).  

However my version lacked chunked encoding and background support.  With the work on the new release, I decided it would be nice to integrate the upstream version by contributing a vanilla XMLHttpRequest version.

Note: It's very minimally tested both by specs and real-world use.  I have upgraded my project to it and tested in all modern browsers, but I may have missed corner cases.


SamSaffron:
I really like the idea of shaking the tight dependency on jQuery, but I am very happy to have a soft dependency. 

Lots of people expect all ajax calls to go via jQuery which gives you pipelines for decorating and so on. 

Ideally the way I see it

MessageBus.ajax = (jQuery && jQuery.ajax) || MessageBus.ajaxImplementation; 

I am super happy to see a test suite added for MessageBus. Also think it would be mighty fine to ship a MessageBus.ajaxImplementation method (probably in its own file, cause you would simply bundle it).

Totally open to also supporting json encoded body, this was raised before and makes interacting with us via other clients a bit easier.

nathanstitt:
@SamSaffron thanks for considering the change!  I've now:
 * restored `MessageBus.ajax` to default to jQuery and added `MessageBus.ajaxImplementation` as a fallback.  
 * Updated MessageBus to use JSON body instead of post parameters.
 * Added a few more client specs.  If there's any additional options that you'd like tested I can take a shot at adding them.

Right now the `XMLHttpRequest` implantation is in a separate file, but I worry that it might be confusing to users to have to include it. The current implementation is only 40 lines so perhaps we should consider moving it into the main `message-bus.js` file?

SamSaffron:
Excellent progress! 

I think the correct thing to do here from a server's side is to support both JSON payload and "the old way", just set `Content-Type: application/json` and have the server switch on that, if it is application/json ... use the json parsing, otherwise use form-data

in theory form-data is ever so slightly more efficient on the server so we should keep using it for the standard jQuery implementation

Regarding the confusion with the extra file, I think it is reasonable to require the extra file as the vast majority already have jQuery in place, what we could do is have it raise an exception if it is unable to resolve a ajax implementation notifying users they need to add the file. 

Overall super happy with this work being done!

nathanstitt:
@SamSaffron Take a look at the current state and see what you think. I've been testing it and seems to work well for me at least.

BTW, I've had some puzzling spec timeouts with Travis. Sometimes a spec run will lockup for 10 minutes until Travis kills it.  There's no pattern to it that I can discern, sometimes it's the Postgresql run, other times Redis.  Is this something you've noticed on `master` as well, or should we be concerned with it on this PR?

SamSaffron:
I am really liking this! 

@jeremyevans  server wise there only a tiny change but worth you reviewing as well. 

Very happy to have all the client side testing! 

The tests failing are mostly due to us pushing the limits of testing here with the forking tests, many projects avoid even touching this kind of stuff, I would like to fix it though.

jeremyevans:
Rack middleware change looks good to me.

SamSaffron:
OK, let's merge this in ... it also resolves https://github.com/SamSaffron/message_bus/pull/42 

Thanks for all your work here @nathanstitt

PR #89, opened 2016-03-29
Add a noConflict method to remove MessageBus from global namespace

nathanstitt:
While adding it I also noticed a few missing `var` declarations and switched the ajax implementation from the previous PR to load after MessageBus.

If there's interest I can also add a UMD definition.  I personally don't need that but it's probably the 'right thing' to have.


SamSaffron:
nice!

SamSaffron:
I guess we can add a UMD .. not against it, but also don't have a use for it atm

PR #96, opened 2016-04-19
MessageBus native ajax implementation added to vendor folder

muratcorlu:
We couldn't import `message-bus-ajax.js` file to our rails project because of it's not reachable with gem package. So I added a symlink to vendor folder for message-bus-ajax.js file.


SamSaffron:
thanks!

Issue #98, opened 2016-04-20
Chat example set header without JQuery  (using message-bus-ajax.js)


ffabreti:
How can I set header like [in chat example...](https://github.com/SamSaffron/message_bus/blob/fc367a176e08d10266805899e38687f7b28850b5/examples/chat/chat.rb#L124-L127)
```
MessageBus.ajax = function(args){
          args["headers"]["X-NAME"] = name;
          return $.ajax(args);
        };
```
... without JQuery support (using message-bus-ajax.js)  ?

tried this without success:
```
window.MessageBus.ajax = function(args){
       args = args || {};
       args["headers"]["X-NAME"] = name;
       return new XMLHttpRequest(args);
 };
```
Also tried: 
```
            window.MessageBus.ajax({
                headers: {
                    'X-name': function() { return currentUser; }
                }
            });
```
but it complains about a lot of other not defined options.

ffabreti:
My problem is that I cannot set headers for `longPoller()` that is defined inside `start()` 
I need to set headers on pooling so that I can send username to backend keep registering user presence.

SamSaffron:
@nathanstitt  any ideas here? do we need to improve the ajax implementation ?

nathanstitt:
Hm, it does read and set headers [from the options](https://github.com/SamSaffron/message_bus/blob/master/assets/message-bus-ajax.js#L22-L24), but there isn't a way to set them exposed from outside MessageBus.  Maybe we should use an extend instead of set [where they're defined]( https://github.com/SamSaffron/message_bus/blob/master/assets/message-bus.js#L123)  ? 

If that sounds like a good idea I can whip it up pretty quickly.

ffabreti:
Thanks Guys. 
Tested and it works great.

PR #99, opened 2016-04-21
Add headers config option

nathanstitt:
Adds a `MessageBus.headers` configuration.  As noted on #98 it was previously impossible set custom headers with requests.


SamSaffron:
I like this change! so tidy

PR #108, opened 2016-05-04
improve readme on how user_id_lookup is used

### LINKED ISSUE #107: improve readme on how is user_id_lookup used
if I understand correctly, this:
```
MessageBus.configure(user_id_lookup: proc do |env|
  # return the user id here
end)
```
is only checked when client pooling reaches the server, and before processing the pooling request.
Correct me if I am wrong!

Been so, Would you accept a pull request for the docs explaining it? Something like:
```
MessageBus.configure(user_id_lookup: proc do |env|
   # this lookup occurs on JS-client poolings, so that server can retrieve backlog 
   # for the client considering/matching/filtering user_id set on published message
   # if user_id is not set on publish time, any user_id returned here will receive the message
 
   # return the user id here
end)
```

ffabreti:
closes #107

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,10 @@ MessageBus.publish "/channel", "hello", client_ids: ["XXX","YYY"]
 # message bus determines the user ids and groups based on env
 
 MessageBus.configure(user_id_lookup: proc do |env|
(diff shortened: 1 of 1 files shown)


SamSaffron:
sure, thanks

PR #110, opened 2016-05-17
Update message-bus.js

washu:
added async: true as an explicit option instead of the default value as Firefox seems to intermittently hang up the UI during a long poll without a default


SamSaffron:
Sure, we can be explicit here, keep in mind the default is async true its just for people that somehow amend the default in jQuery.

washu:
it shouldnt

PR #114, opened 2016-06-08
adding an extra example to use with rails/devise

joseluistorres:
While starting to work with message_bus I struggled to find the right way to get the current user ids, I found this issue https://github.com/SamSaffron/message_bus/issues/64 and just wanted to share it with new-comers.

Thanks for this tool!!! 👍


SamSaffron:
cool! thanks!

PR #117, opened 2016-06-25
Add a status function

danweller18:
Adding status function is useful to determine where to call MessageBus.start() & MessageBus.stop()

### DIFF
--- a/assets/message-bus.js
+++ b/assets/message-bus.js
@@ -303,9 +303,9 @@
     },
 
     status: function() {
-      if (started) {
+      if (started == true) {
         return "Started";
-      } else if (stopped) {
+      } else if (stopped == true) {
         return "Not Started";
       } else {
         console.log("Error cannot determine status. Please restart.");
--- a/assets/message-bus.js
+++ b/assets/message-bus.js
@@ -302,6 +302,17 @@
       later = [];
     },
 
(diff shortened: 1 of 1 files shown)


danweller18:
I accidentally uploaded the wrong version first. The second commit works

SamSaffron:
hmm don't we already have this in .diagnostics? https://github.com/danweller18/message_bus/blob/aea43952ae79cb09daa7ede5e4fbc246ebc99da4/assets/message-bus.js#L287-L293

danweller18:
hmm I didn't realize diagnostics logged the status. I still want to get a check of the current state of MessageBus. I am using it to turn on if MessageBus is off in my javascript. I cannot think of a way to do that by using just diagnostics

SamSaffron:
I think there are 3 states .. started, stopped and paused ... if you amend to return the 3 states will be happy to merge.

SamSaffron:
I still prefer a discrete "status" function cause stuff with console.log is not that friendly, perhaps go with original just return 1 of the 3 states as a string.

SamSaffron:
this is implemented now per 

https://github.com/SamSaffron/message_bus/commit/9d78a06c2f4d3ea8d1a79da3af5c5bfa0c3a0cfc

Thanks!

PR #121, opened 2016-06-30
Changed MessageBus message form a struct to an object.

washu:
Change MessageBus to a class instead of a Struct, I added some backwards compat mthod for struct member access.
Done so that subscribers threads in development where class reloading is occurig in a seperate thread get module load/unload error about version.

### DIFF
--- a/lib/message_bus/message.rb
+++ b/lib/message_bus/message.rb
@@ -1,7 +1,14 @@
-class MessageBus::Message < Struct.new(:global_id, :message_id, :channel , :data)
-
-  attr_accessor :site_id, :user_ids, :group_ids, :client_ids
-
+class MessageBus::Message
+  INDEX_MAP={0=> :global_id, 1=> :message_id, 2=> :channel, 3=> :data}
+  attr_accessor :site_id, :user_ids, :group_ids, :client_ids, :global_id,:message_id,:channel,:data
+  
+  def initialize(_gi,_mi,_ch,_da)
+    @global_id = _gi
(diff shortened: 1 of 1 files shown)


jeremyevans:
I'm against this change.  You shouldn't be reloading just by loading the files on top of the existing code.  You need to remove the `MessageBus::Message` constant, then load the file.  Also, I'm not sure why you would be reloading this code in development. In general reloading is specific to application code, it doesn't reload all files in all gems.

The pull request message implies that this change will cause an error to occur where an error does not occur now.  I don't think that's true, I think this stops an existing superclass mismatch error from happening.

In terms of a technical reason to be against this change, this slows down instantiation by about 22%, and `#[]` by 75%.

washu:
reloading is normal is rails development,  every request reloads classes

as for speed i havent run any type of benchmark

jeremyevans:
I don't believe Rails reloads all files from gems, and certainly if you aren't modifying the file, it shouldn't be reloading it.  However, it's been a few years since I last used Rails.  If this was default behavior in Rails, I assume other people would be experiencing the issue on a regular basis.  Anyway, I've said my piece, it's up to @SamSaffron to make the decision on whether this should go in.

washu:
it maybe due to my use case, i created a some static module function to help setup the calls around and it may cause a refrence handle or something on reload .. its loading from the lib directory

washu:
closed and revoked more debugging found root cause and it wasnt the struct itself to removing pull request

Issue #70, opened 2016-02-09
getting a 404 on the polling


karmajunkie:
Everything works in development when I run with unicorn, without the addition to my unicorn.rb recommended in the readme. However, in production-like circumstances (our staging server) I'm seeing extremely long waits before a published message hits the client.  I'm running Rails 4.2 with the master branch of message_bus.

So, in an effort to get to the bottom of that, I made the addition to the unicorn configuration file, and immediately started having trouble. The first problem was that Rails complained about missing the `message-bus.js` asset, which I worked around by copying it out of the gem. Now I'm getting 404's on the polling requests. The output of `rake middleware` is as follows:

```
use Honeybadger::Rack::UserInformer
use Honeybadger::Rack::UserFeedback
use Honeybadger::Rack::ErrorNotifier
use Rack::Sendfile
use ActionDispatch::Static
use #<ActiveSupport::Cache::Strategy::LocalCache::Middleware:0x007fb3ec49fa50>
use Rack::Runtime
use Rack::MethodOverride
use ActionDispatch::RequestId
use Rails::Rack::Logger
use ActionDispatch::ShowExceptions
use ActionDispatch::DebugExceptions
use BetterErrors::Middleware
use ActionDispatch::RemoteIp
use ActionDispatch::Reloader
use ActionDispatch::Callbacks
use ActiveRecord::Migration::CheckPending
use ActiveRecord::ConnectionAdapters::ConnectionManagement
use ActionDispatch::Cookies
use ActionDispatch::Session::CookieStore
use MessageBus::Rack::Middleware
use ActionDispatch::Flash
use ActionDispatch::ParamsParser
use Rack::Head
use Rack::ConditionalGet
use Rack::ETag
use Warden::Manager
run MyApp::Application.routes
```

Any ideas of what I'm doing wrong here?

SamSaffron:
are you running in a subfolder? 

Can you try bumping MessageBus::Rack::Middleware up to position 0?

karmajunkie:
Not sure what you mean by running in a subfolder—the application is run via unicorn from the application root, with the config in `config/unicorn.rb`.

[...]

PR #126, opened 2016-08-11
[ci skip] Update puma instructions for Rails 5

dleavitt:
Requiring message_bus in the puma config, under Rails 5, when running via the `puma` command (but not via `rails s`) causes message_bus to be loaded before the rest of the Rails stack, which keeps it from initializing correctly.

I believe this line is the cause of #70 

You generally won't need to explicitly require it under rails anyway. If it's in your Gemfile it should be required by the time the `after_fork` hook is called.


SamSaffron:
weird, sure seems fine

PR #133, opened 2017-03-15
Better logging when MessageBus drops old message.

tgxworld:


SamSaffron:
cool, thanks

need to fix travis though (but unrelated)

PR #137, opened 2017-04-04
Update middleware.rb

washu:
convert all constant strings, to froze const vars.
performance improvement for the gc cycles. it also uses ruby 2.4 match? for faster regex tests if available

### DIFF
--- a/lib/message_bus/rack/middleware.rb
+++ b/lib/message_bus/rack/middleware.rb
@@ -56,25 +56,66 @@ class MessageBus::Rack::Middleware
     end.to_a
     JSON.dump(m)
   end
-
+  # GC helping.
+  PATH_INFO = 'PATH_INFO'.freeze
+  CONTENT_TYPE = 'Content-Type'.freeze
+  SLASH = '/'.freeze
+  BUS_REGEX = /^\/message-bus\//.freeze
+  USE_MATCH = BUS_REGEX.respond_to?(:match?)
+  BROADCAST = '/message-bus/broadcast'.freeze
+  DIAGNOSTICS = '/message-bus/_diagnostics'.freeze
+  CHANNEL = 'channel'.freeze
+  DATA = 'data'.freeze
+  TEXT_HTML = 'text/html'.freeze
+  SENT = 'sent'.freeze
+  NOT_FOUND = 'not found'.freeze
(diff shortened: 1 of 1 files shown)


lib/message_bus/rack/middleware.rb
```diff
@@ -56,25 +56,66 @@ def self.backlog_to_json(backlog)
...
+  NO_SNIFF = 'nosniff'.freeze
+  TRANSFER_ENCODING = 'Transfer-Encoding'.freeze
+  CHUNKED = 'chunked'.freeze
+  TEXT_CONTENT_TYPE_LINE = 'text/plain; charset=utf-8'.freeze
+  EMPTY_LIST = "[]".freeze
+  
   def call(env)
 
-    return @app.call(env) unless env['PATH_INFO'] =~ /^\/message-bus\//
+    return @app.call(env) unless USE_MATCH ? BUS_REGEX.match?(env[PATH_INFO]) : env[PATH_INFO] =~ BUS_REGEX
```
SamSaffron: hmmm no regex is even needed here, can simply use `start_with?`

lib/message_bus/rack/middleware.rb
```diff
@@ -107,27 +148,27 @@ def call(env)
...
     end
 
-    if env["REQUEST_METHOD"] == "OPTIONS"
-      return [200, headers, ["OK"]]
+    if env[REQUEST_METHOD] == OPTIONS
+      return [200, headers, [OK]]
     end
 
     long_polling = @bus.long_polling_enabled? &&
-                   env['QUERY_STRING'] !~ /dlp=t/.freeze &&
```
SamSaffron: no need for regex here as well.

SamSaffron:
Overall I feel this makes the code a bit less clear, instead perhaps add 

```
# frozen_string_literal: true
```

Then we only need constants for stuff that is used more than once, the indirection introduces some complexity.

washu:
frozen_string_literal is a ruby 2.3+ feature only and is a pragma directive

SamSaffron:
agree, but I think it is fair to provide this perf boost only to 2.3+

washu:
in that case i can abandon the pull request, if you want to just go ahad and add the 3 lines that would be left. the 2 starts with and the comment

Issue #124, opened 2016-08-01
No such middleware to insert before: ActionDispatch::Flash


tcoenraad:
Similar to #24, I get after adding `message_bus` on a clean Rails 5 API project the following error:

```sh
$ bundle exec rails middleware                                                                                                                                                                                                                                                                                                                             
rails aborted!
No such middleware to insert before: ActionDispatch::Flash
./rails5/config/environment.rb:5:in `<top (required)>'
./rails5/bin/rails:9:in `require'
./rails5/bin/rails:9:in `<top (required)>'
./rails5/bin/spring:13:in `require'
./rails5/bin/spring:13:in `<top (required)>'
bin/rails:3:in `load'
bin/rails:3:in `<main>'
Tasks: TOP => middleware => environment
(See full trace by running task with --trace)
```

The `begin` `rescue` block on line https://github.com/SamSaffron/message_bus/blob/master/lib/message_bus/rails/railtie.rb#L14 is somehow not triggered. When I alter this particular line in a similar way as done in https://github.com/SamSaffron/message_bus/pull/35/files `message_bus` works perfectly.

Can you look into this?

SamSaffron:
hmmm maybe this just needs a rescue => e ?

kubakrzempek:
@SamSaffron it is not that easy...

The thing is the middleware config in Rails app operates on a [MiddlewareStackProxy](http://api.rubyonrails.org/classes/Rails/Configuration/MiddlewareStackProxy.html) which more or less acts as a recorder of the operations that is applied over default stack when the app is initialized.

That means you can `app.middleware.insert_before(ActionDispatch::Flash, MessageBus::Rack::Middleware)` even though the `ActionDispatch::Flash` is deleted already and this won't raise any error because at that point nothing is being set up just yet. If you want to catch the potential error you can do so but here:
```rb
# config/environment.rb
# Load the Rails application.
require File.expand_path('../application', __FILE__)

# Initialize the Rails application.
begin
  Rails.application.initialize!
rescue
  # but most likely you won't be able to alter the middleware stack here
end
```

Push the `MessageBus::Rack::Middleware` at the bottom MIGHT be a solution, but from what I've seen in #35 you are not going to support that solution, so I'd give the user a chance to decide where to push it in their stack so the user take the responsibility.

I'll come up with a PR with my proposal on how to solve that issue (I really want to use neither `ActionDispatch::Flash` nor a forked version of the lib... :)).

PR #140, opened 2017-05-23
Handle missing ActionDispatch::Flash in Railtie

kubakrzempek:
The middleware config in Rails app operates on a MiddlewareStackProxy which more or less acts as a recorder of the operations that is applied over default stack when the app is initialized, hence a regular `rescue` won't work here.

There are two cases to cover:
1. Someone already "deleted" (meaning it will be deleted in the end but it is not just yet)
   ActionDispatch::Flash from the stack
2. ActionDispatch::Flash was never in the stack

No. 1 happens with Rails4 apps that are configured to be API-only or if someone doesn't want the middleware in their stack.
No. 2 is a Rails5 API-only app behaviour.

To handle either case the initializer in `MessageBus::Rails::Railtie` is looking for both cases and applies the MessageBus middleware appropriately. The default behaviour can be easily changed via regular Rails config block.

```ruby
 # config/initializers/message_bus.rb
Rails.application.config do |config|
  # do anything you wish with config.middleware here
end
```

The config was tested on Rails4, Rails4 API-only, Rails5 and Rails5 API-only.
It should solve #124

### DIFF
--- a/lib/message_bus/rails/railtie.rb
+++ b/lib/message_bus/rails/railtie.rb
@@ -11,12 +11,24 @@ class MessageBus::Rails::Railtie < ::Rails::Railtie
     #
     # To handle either case, we insert it before ActionDispatch::Flash.
     #
-    begin
-      app.middleware.insert_before(ActionDispatch::Flash, MessageBus::Rack::Middleware)
-    rescue
+    # For APIs or apps that have ActionDispatch::Flash deleted from the middleware
+    # stack we just push MessageBus to the bottom.
+    if api_only?(app.config) || flash_middleware_deleted?(app.middleware)
       app.middleware.use(MessageBus::Rack::Middleware)
+    else
+      app.middleware.insert_before(ActionDispatch::Flash, MessageBus::Rack::Middleware)
     end
 
     MessageBus.logger = Rails.logger
+  end
+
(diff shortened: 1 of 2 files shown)


SamSaffron:
awesome, test failure is unrelated

lib/message_bus/rails/railtie.rb
```diff
@@ -11,12 +11,24 @@ class MessageBus::Rails::Railtie < ::Rails::Railtie
...
     MessageBus.logger = Rails.logger
   end
+
+  def api_only?(config)
+    return false unless config.respond_to?(:api_only)
+    config.api_only
+  end
+
+  def flash_middleware_deleted?(middleware)
+    ops = middleware.instance_variable_get(:@operations)
```
aCandidMind: For extra safety (this is using private internals) `|| []` should be added.

SamSaffron:
Sadly the crawl no longer works on Rails 6.0.3 

I am going to have to remove it

https://github.com/rails/rails/commit/fedde239dcee256b417dc9bcfe5fef603bf0d952#diff-533a9a9cc17a8a899cb830626089e5f9

I think the best thing here is just to remove the check altogether cause given the new implementation by @kamipo this simply can not work, going to do a new release without this.

PR #141, opened 2017-06-21
Add performance logging

booty:
ref: https://www.pivotaltracker.com/story/show/147226655

The intent is to add performace logging to message_bus so that we can see why/where it's slow.

**Note For Reviewers**

It looks like there's more flux in this PR than there truly is. Note that most of the changed lines are actually have just had their indent level increased.

### DIFF
--- a/lib/message_bus/backends/redis.rb
+++ b/lib/message_bus/backends/redis.rb
@@ -81,44 +81,54 @@ class MessageBus::Redis::ReliablePubSub
   end
 
   def publish(channel, data, queue_in_memory=true)
-    redis = pub_redis
-    backlog_id_key = backlog_id_key(channel)
-    backlog_key = backlog_key(channel)
-
-    global_id = nil
+    log_data = {}
     backlog_id = nil
+    log_data[:publish_seconds] = Benchmark.realtime do
+      redis = pub_redis
+      backlog_id_key = backlog_id_key(channel)
+      backlog_key = backlog_key(channel)
 
-    redis.multi do |m|
-      global_id = m.incr(global_id_key)
-      backlog_id = m.incr(backlog_id_key)
-    end
+      global_id = nil
+      backlog_id = nil
 
-    global_id = global_id.value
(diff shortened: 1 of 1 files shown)


SamSaffron:
The link to the story is not working, can you try linking again?

SamSaffron:
Note, I would strongly prefer not to make this default behavior, instead add a new backend that takes in an existing backend in the constructor and proxies and measures. 

That way people can use this with the PG provider if they wish and measuring code is not mixed in with the default redis backend

booty:
@SamSaffron I'm sorry to bother you with this. This was meant to be a PR against our fork of `message_bus`, not your primary `message_bus` repo. My face is very red right now.

This PR definitely uglifies things and is not suitable for public consumption; it's meant to be something we'll use internally to gain insight into some perf issues.

SamSaffron:
I definitely want more info here, email me privately if you wish with any of your findings. 

What kind of scale are you pushing through the bus?

booty:
I appreciate it, @SamSaffron! As a former dapper user, the number of beverages I owe you is growing alarmingly.

We have a cluster of three (I believe?) redis servers. 

According to NewRelic, we push about 20,000 calls per minute to `zrangebyscore` and message_bus's `backlog` and `global_backlog` are the only instances of `zrangebyscore` in our codebase. Avg. execution time of `zrangebyscore` (which includes network overhead, I think) is under 10ms, but sometimes spikes up over 1,000ms (std dev = 15ms during peak hours, FWIW)

I'm not sure of our write throughput. I believe it's not more than a few hundred messages per minute, but I'll have a more exact number after we get this quick-and-dirty-perf-logging branch of message_bus out to production.

Global backlog max is 2,000 entries and backlog max is 1,000 entries.

It seems very possible that we're pushing message_bus beyond what it's meant to do! Or, the redis backend, at least. I suspect the pg backend would not have this issue, but postgres isn't an option for us at the moment.

I'll get you more concrete numbers when we have them.

Issue #146, opened 2017-08-03
Errno::EPIPE message_bus/client.rb in write at line 156


cannikin:
I see a lot of these in my logs...is this just the result of someone closing their browser or navigating away before MessageBus gets a chance to tell them something?

Ruby 2.4.1
Rails 5.1.2
Puma 3.8.2

```
Errno::EPIPE: Broken pipe
  from message_bus/client.rb:156:in `write'
  from message_bus/client.rb:156:in `write_headers'
  from message_bus/client.rb:175:in `write_chunk'
  from message_bus/client.rb:48:in `ensure_first_chunk_sent'
  from message_bus/rack/middleware.rb:149:in `block in call'
  from message_bus/client.rb:20:in `block in synchronize'
  from message_bus/client.rb:20:in `synchronize'
  from message_bus/client.rb:20:in `synchronize'
  from message_bus/rack/middleware.rb:146:in `call'
  from rack/session/abstract/id.rb:232:in `context'
  from rack/session/abstract/id.rb:226:in `call'
  from action_dispatch/middleware/cookies.rb:613:in `call'
  from action_dispatch/middleware/callbacks.rb:26:in `block in call'
  from active_support/callbacks.rb:97:in `run_callbacks'
  from action_dispatch/middleware/callbacks.rb:24:in `call'
  from action_dispatch/middleware/debug_exceptions.rb:59:in `call'
  from action_dispatch/middleware/show_exceptions.rb:31:in `call'
  from rails/rack/logger.rb:36:in `call_app'
  from rails/rack/logger.rb:24:in `block in call'
  from active_support/tagged_logging.rb:69:in `block in tagged'
  from active_support/tagged_logging.rb:26:in `tagged'
  from active_support/tagged_logging.rb:69:in `tagged'
  from rails/rack/logger.rb:24:in `call'
  from action_dispatch/middleware/remote_ip.rb:79:in `call'
  from request_store/middleware.rb:9:in `call'
  from action_dispatch/middleware/request_id.rb:25:in `call'
  from rack/method_override.rb:22:in `call'
  from rack/runtime.rb:22:in `call'
  from active_support/cache/strategy/local_cache_middleware.rb:27:in `call'
  from action_dispatch/middleware/executor.rb:12:in `call'
  from rack/sendfile.rb:111:in `call'
  from action_dispatch/middleware/ssl.rb:66:in `call'
  from raven/integrations/rack.rb:50:in `call'
[...]

PR #147, opened 2017-08-11
Rescue from Errno::EPIPE when sending a message

cannikin:
For issue #146

### DIFF
--- a/lib/message_bus/client.rb
+++ b/lib/message_bus/client.rb
@@ -98,6 +98,9 @@ class MessageBus::Client
     else
       write_and_close json
     end
+  rescue Errno::EPIPE => e
+    # client went away
+    @io = nil
   end
 
   def subscriptions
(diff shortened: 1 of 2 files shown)


SamSaffron:
@jeremyevans  I am a tiny bit mixed on this, agree that eating the error is probably best, what do you think?

jeremyevans:
I agree that removing disconnected clients should probably not raise errors, but handling just EPIPE seems questionable (though it may work fine for puma).  Maybe rescuing IOError and SystemCallError?

SamSaffron:
Yeah, a wider net sounds like a good idea

cannikin:
Updated! Do you guys still like the spec that's there or should I try to raise anything else to verify it'll get caught?

SamSaffron:
I think this is fine, let's try it out.

PR #148, opened 2017-08-15
Moves error handling to `write_headers` instead

cannikin:
See my comment at the end of #146 [here](https://github.com/SamSaffron/message_bus/issues/146#issuecomment-322351262). I'm seeing the `Errno::EPIPE` much less frequently, but it still pops up from a couple other places that make calls to write outside of `<<`.

This PR moves the error handling to `write_headers` which is called from everywhere else that's been raising the error, including `<<`. I now have `write_headers` return a boolean of whether it was successful in writing those headers, and the methods that call `write_headers` check that it returned `true` before writing the rest of their response.

The only hiccup I came across was trying to make a one-liner in the check in `write_and_close`: my test didn't have an `async_response` object to give to the client, so an error was raised when the check looked like `if @io && write_headers`:

```
NoMethodError: undefined method `<<' for nil:NilClass
    /Users/rob/Sites/gems/message_bus/lib/message_bus/client.rb:221:in `write_and_close'
    /Users/rob/Sites/gems/message_bus/lib/message_bus/client.rb:99:in `<<'
    spec/lib/message_bus/client_spec.rb:56:in `block (3 levels) in <top (required)>'
```

I found an example of setting `async_response` in the ConnectionManager spec but I don't feel like I should be crossing concerns and trying to add fake responses into this Client spec. Could I stub `async_response` in the Client spec to respond to `<<` and `done`?

Or maybe it's safe to leave that as two separate checks and only have `@async_response` set if `@io` is initially `nil` and not if it becomes `nil` as a result of `write_headers` failing? Could a message get lost?

### DIFF
--- a/lib/message_bus/client.rb
+++ b/lib/message_bus/client.rb
@@ -98,9 +98,6 @@ class MessageBus::Client
     else
       write_and_close json
     end
-  rescue IOError, SystemCallError
-    # client went away
-    @io = nil
   end
 
   def subscriptions
@@ -170,12 +167,16 @@ class MessageBus::Client
     else
       @io.write(TYPE_JSON)
     end
+    true
+  rescue IOError, SystemCallError
+    # client went away
+    @io = nil
+    false
   end
 
   def write_chunk(data)
     @bus.logger.debug "Delivering messages #{data} to client #{client_id} for user #{user_id} (chunked)"
-    if @io && !@wrote_headers
-      write_headers
+    if @io && !@wrote_headers && write_headers
       @io.write(CHUNKED_ENCODING)
       # this is required otherwise chrome will delay onprogress calls
       @io.write(NO_SNIFF)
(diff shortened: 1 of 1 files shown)


SamSaffron:
hmmm ... implementation wise maybe we should be more consistent. 

We handle this here: 

https://github.com/SamSaffron/message_bus/blob/master/lib/message_bus/connection_manager.rb#L26-L32


and here:

https://github.com/SamSaffron/message_bus/blob/master/lib/message_bus/rack/middleware.rb#L207-L214

Maybe we are just missing another spot? Maybe we can just pull out the handling from client.rb and ensure stuff that attempts to send messages always catches.

cannikin:
Every one of the errors has this block of code as the "lowest common denominator" in the backtrace: https://github.com/SamSaffron/message_bus/blob/master/lib/message_bus/rack/middleware.rb#L147-L151

```ruby
client.synchronize do
  client.deliver_backlog(backlog)
  add_client_with_timeout(client)
  client.ensure_first_chunk_sent
end
```

We're using long polling and have rack.hijack so that's the only block we enter, but there are a couple of `client.synchronize` blocks nearby by that have similar logic. 

Is it safe to rescue in the middleware and just return a static response or could a message get dropped? That particular block returns a 418(!) right after the `synchronize`. The last `else` in the chain returns `[200, headers, ["[]"]]`.

I do see an undocumented feature at the end of the method there:

```ruby
rescue => e
  if @bus.on_middleware_error && result = @bus.on_middleware_error.call(env, e)
    result
  else
    raise
  end
end
```

So just defining an `on_middleware_error` block in my config will allow me to do whatever I want with the error, including just discarding it?

Maybe my PR should just be adding documentation to *this* feature instead of swallowing the error? :)

SamSaffron:
> Maybe my PR should just be adding documentation to this feature instead of swallowing the error? :) 

I support that. Maybe just do that? and revert the original rescue. feels somewhat cleaner

cannikin:
Sounds good! I'll close this one and open a new one.

PR #150, opened 2017-08-16
Fix broken `MessageBus.redis_config`.

tgxworld:
cc @SamSaffron 
 Introduced in https://github.com/tgxworld/message_bus/commit/a08353c59a6754938691de27d1cd86353841698b


SamSaffron:
good catch!

PR #151, opened 2017-08-16
FIX: Publish status update to all subscribed clients when a message i…

tgxworld:
…s filtered.

### DIFF
--- a/lib/message_bus/client.rb
+++ b/lib/message_bus/client.rb
@@ -121,21 +121,31 @@ class MessageBus::Client
 
   def backlog
     r = []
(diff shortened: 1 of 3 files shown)


tgxworld:
Spec failures does not seem to be related

SamSaffron:
this looks good! awesome

benlangfeld:
What's the purpose of this change? I'm not sure I understand what it achieves.

SamSaffron:
This stops situation where a client keeps asking for messages they clearly have no access to forcing the server to a whole bunch of work on every single call.

PR #152, opened 2017-08-23
don't start message bus in rake tasks

mikz:
* turn off message bus by rails railtie when running rake
* don't initialize middleware subscribed when off
* method to detect if message bus is off

Otherwise running MessageBus with PostgreSQL backend will break tasks like dropping database etc.

As there is always opened db connection.

### DIFF
--- a/lib/message_bus/rails/railtie.rb
+++ b/lib/message_bus/rails/railtie.rb
@@ -23,6 +23,10 @@ class MessageBus::Rails::Railtie < ::Rails::Railtie
     MessageBus.logger = Rails.logger
   end
 
+  rake_tasks do
+    MessageBus.off
+  end
+
   def api_only?(config)
     return false unless config.respond_to?(:api_only)
     config.api_only
--- a/lib/message_bus.rb
+++ b/lib/message_bus.rb
@@ -116,6 +116,10 @@ module MessageBus::Implementation
     @config[:long_polling_interval] || 25 * 1000
   end
 
+  def off?
+    @off
+  end
+
   def off
     @off = true
   end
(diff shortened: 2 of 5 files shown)


SamSaffron:
I definitely want to fix `off`, but this is going to have to be an opt-in thing, we have many rake tasks that depend on having a functional message bus.

mikz:
@SamSaffron ok, that makes sense. 

Does using Rails OrderedOptions to have `config.message_bus.` sounds good?

SamSaffron:
any news on this?

mikz:
@SamSaffron Spent some time with it, but it is quite hard to make it work with spring. 

The main problem is that the middleware is initialized early - when spring is preloading the application.

There are two problems
 - spring is keeping a live copy with background threads etc. 
 - `rake_tasks` is evaluated after initializing all middleware (which already creates the thread)

lib/message_bus/rails/railtie.rb
```diff
@@ -23,6 +23,10 @@ class MessageBus::Rails::Railtie < ::Rails::Railtie
     MessageBus.logger = Rails.logger
   end
 
+  rake_tasks do
```
benlangfeld: @SamSaffron If this block were removed, would you accept the PR?

benlangfeld:
My use case for this is that I have an application build which needs to execute rake tasks in an environment where Redis does not yet exist (bootstrapping a full environment on Kubernetes). I have publications in Rake tasks, but not subscriptions, so this would be perfect.

SamSaffron:
@benlangfeld yes totally having a proper api to turn the bus off is absolutely fine. turning off for rake tasks though has got to be an opt in thing.

benlangfeld:
@mikz Any objection to me re-submitting this as an opt-in thing?

benlangfeld:
Closing in favour of https://github.com/SamSaffron/message_bus/pull/185

PR #155, opened 2017-11-08
Refactor MessageBus::Message.decode to reduce object allocations

jamescook:
I changed the `decode` method to use positional lookups instead of allocating Range objects. Using ruby 2.4 on my macbook yields a ~25% performance improvement.

```
#benchmark-ips
Warming up --------------------------------------
             decode1    37.131k i/100ms
             decode2    45.849k i/100ms
Calculating -------------------------------------
             decode1    412.939k (± 5.2%) i/s -      2.079M in   5.050547s
             decode2    522.827k (± 3.7%) i/s -      2.613M in   5.005663s

Comparison:
             decode2:   522827.1 i/s
             decode1:   412938.7 i/s - 1.27x  slower

# benchmark-memory
Calculating -------------------------------------
             decode1   432.000  memsize (     0.000  retained)
                        10.000  objects (     0.000  retained)
                         3.000  strings (     0.000  retained)
             decode2   272.000  memsize (     0.000  retained)
                         6.000  objects (     0.000  retained)
                         4.000  strings (     0.000  retained)

Comparison:
             decode2:        272 allocated
             decode1:        432 allocated - 1.59x more
```

[Benchmark script](https://gist.github.com/jamescook/f00d0850d7df5bf926fefbf461a14850)

### DIFF
--- a/lib/message_bus/message.rb
+++ b/lib/message_bus/message.rb
@@ -11,7 +11,8 @@ class MessageBus::Message < Struct.new(:global_id, :message_id, :channel , :data
 
     global_id  = encoded[0, s1+1].to_i
     message_id = encoded[(s1 + 1), (s2 - s1 - 1)].to_i
-    channel    = encoded[(s2 + 1), (s3 - s2 - 1)].gsub("$$123$$", "|")
+    channel    = encoded[(s2 + 1), (s3 - s2 - 1)]
+    channel.gsub!("$$123$$", "|")
     data       = encoded[(s3 + 1), encoded.size]
 
     MessageBus::Message.new(global_id, message_id, channel, data)
--- a/lib/message_bus/message.rb
+++ b/lib/message_bus/message.rb
@@ -9,8 +9,12 @@ class MessageBus::Message < Struct.new(:global_id, :message_id, :channel , :data
     s2 = encoded.index("|", s1 + 1)
     s3 = encoded.index("|", s2 + 1)
 
-    MessageBus::Message.new(encoded[0..s1].to_i, encoded[(s1 + 1)..s2].to_i,
-                            encoded[(s2 + 1)..(s3 - 1)].gsub("$$123$$", "|"), encoded[(s3 + 1)..-1])
+    global_id  = encoded[0, s1+1].to_i
+    message_id = encoded[(s1 + 1), (s2 - s1 - 1)].to_i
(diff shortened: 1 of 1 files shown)


lib/message_bus/message.rb
```diff
@@ -9,8 +9,12 @@ def self.decode(encoded)
     s2 = encoded.index("|", s1 + 1)
     s3 = encoded.index("|", s2 + 1)
 
-    MessageBus::Message.new(encoded[0..s1].to_i, encoded[(s1 + 1)..s2].to_i,
-                            encoded[(s2 + 1)..(s3 - 1)].gsub("$$123$$", "|"), encoded[(s3 + 1)..-1])
+    global_id  = encoded[0, s1+1].to_i
+    message_id = encoded[(s1 + 1), (s2 - s1 - 1)].to_i
+    channel    = encoded[(s2 + 1), (s3 - s2 - 1)].gsub("$$123$$", "|")
```
SamSaffron: cool, do we just gsub! here? one less string alloc.

jamescook: @SamSaffron `gsub!` returns nil if there's no match, whereas `gsub` returns the original string. So, it doesn't seem correct (also it breaks a few tests).

SamSaffron: Sorry I mean:

```
channel    = encoded[(s2 + 1), (s3 - s2 - 1)]
channel.gsub!("$$123$$", "|")
```

Its one less string alloc ... calling gsub when there is no hit is the same as calling dup, one extra string

SamSaffron:
do you mind fixing travis, I know it is unrelated, but hard to triple check this is working without it

message_bus.gemspec
```diff
@@ -18,6 +18,6 @@ Gem::Specification.new do |gem|
   gem.version       = MessageBus::VERSION
   gem.required_ruby_version = ">= 1.9.3"
   gem.add_runtime_dependency 'rack', '>= 1.1.3'
-  gem.add_development_dependency 'redis'
+  gem.add_development_dependency 'redis', '3.3.5'
```
SamSaffron: ouch, no we can't do this, unfortunately we are cross comapatible with redis, we work with 4.0.1 as well ... revert this for now I will figure out why it is giving me grief.

jamescook: No problem. I wasn't really sure about it either. Earlier I tried updating .travis.yml to downgrade bundler to 1.15, but it didn't help.

SamSaffron:
thanks heaps for the change!

PR #158, opened 2018-03-09
Fix using global MessageBus, always use the related instance

jeremyevans:
If you are using MessageBus::Instance to create separate instances,
none of that code should call the singleton MessageBus instance.
This fixes all cases where the singleton was used.

In order to implement this, MessageBus::Diagnostics.enable needs
to take an argument for the bus to use.  To retain backwards
compatibility, this argument defaults to MessageBus.


SamSaffron:
Nice, I really like this untangling!

PR #165, opened 2018-10-06
Make minPollInterval customizable

stevenmichaelthomas:
If MessageBus receives very high frequency data (~200 events per second in our current use case), _a lot_ of connections get open since the default minimum interval is only 100ms. 

Being able to modify the `minCallbackInterval` allows us to receive fewer requests with larger payloads, causing less strain on the client by minimizing request volume. 

Let me know if our use case is an outlier, but it seems like a relatively harmless change to make this value configurable. 

I was hoping to write better tests, but given the number of stubs that would be required to get this line to execute, I focused tests on ensuring the default is respected, and overridable.

cc: @jmreid

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,6 +238,7 @@ Setting|Default|Info
 enableLongPolling|true|Allow long-polling (provided it is enable by the server)
 callbackInterval|15000|Safeguard to ensure background polling does not exceed this interval (in milliseconds)
 backgroundCallbackInterval|60000|Interval to poll when long polling is disabled (either explicitly or due to browser being in background)
+minPollInterval|100|When polling requests succeed, this is the minimum amount of time to wait before making the next request.
 maxPollInterval|180000|If request to the server start failing, MessageBus will backoff, this is the upper limit of the backoff.
 alwaysLongPoll|false|For debugging you may want to disable the "is browser in background" check and always long-poll
 baseUrl|/|If message bus is mounted in a subdirectory of different domain, you may configure it to perform requests there
--- a/assets/message-bus.js
+++ b/assets/message-bus.js
@@ -278,7 +278,7 @@
         var interval;
         try {
           if (gotData || aborted) {
-            interval = 100;
+            interval = me.minPollInterval;
           } else {
             interval = me.callbackInterval;
             if (failCount > 2) {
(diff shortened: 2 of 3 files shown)


stevenmichaelthomas:
@SamSaffron let me know if I've missed anything (first contribution to this repo)

SamSaffron:
No probs, it is my weekend will get to this on Monday

SamSaffron:
I am totally not against adding this setting.

We only poll on last resort, usually we would just use chunked encoding and stream messages, is there a reason this is not working? Are you not long polling for some reason? 

Can you add some documentation about the setting to the readme?

stevenmichaelthomas:
I’ll definitely add some documentation. Given the volume of data that we
are passing, and how it is represented visually, we were seeing some
intermittent failures with the chunked requests - because the data is being
used for real-time data visualization, the intermittent failure ends up
looking really janky on the front end as the subsequent successful request
“catches up”.

I’ll share the project as soon as it goes live, but right now we are
disabling chunked encoding to ensure a consistent stream of data (at the
cost of more requests).

Let me know if this makes sense!

SamSaffron:
Interesting! Totally fine with the PR if you are running http2 the saving
from chunked encoding are not as enormous

Curious though is the jankyness browser specific or across all browsers? Do
you have nginx or some proxy in the middle buffering?

stevenmichaelthomas:
We don't have a proxy in this case - I think what we were seeing would normally be fine - it's actually just a request timeout, and message_bus provides great mechanisms to ensure no data is lost, so on the next successful request it catches up, but since we're basically using it to supply rendering data for our visualization, it looks janky. 

All that to say is that I think in this case message bus is working exactly as you'd expect and want, but our use case is weird 😄

SamSaffron:
Thanks, looking forward to see the project, let me know once it is live.

Merging it in now, but we have one more change in the pipeline before we do a version bump (we are pushing distributed cache from Discourse into message bus (optional require) cause we fine we reuse it)

stevenmichaelthomas:
Thanks!

stevenmichaelthomas:
Hey @SamSaffron heads up our project is now live! You can check it out at https://datastories.shopify.com

It has worked like a charm throughout the weekend 😄

SamSaffron:
awesome, just checked out the site and it looks really impressive.

PR #166, opened 2018-10-11
`MesssageBus.publish` accepts `site_id` to explicitly publish to site_id

ramontayag:
Besides relying on `site_id_lookup`, one can also pass `site_id` in `MessageBus.publish` to publish to a specific site.


ramontayag:
Travis is failing in one version of Ruby in a spec that seems unrelated to the changes I made.

SamSaffron:
Change looks A-OK! thanks!

PR #167, opened 2018-10-12
Refactor MessageBus.publish

ramontayag:
- Use keyword arguments to lessen code in method
- Do not go beyond 80 characters wide


SamSaffron:
I appreciate this effort but I am sadly going to decline this, this makes publish every so slightly slower cause it needs to do more work, currently opts is 100% optional and if you pass it nil then there is no extra work added. additionally I worry a lot about subtle regressions here and the existing code is very battle tested to say the least.

ramontayag:
Ah, I wasn't aware of that. No problem!

PR #168, opened 2018-10-12
Add distributed_cache

nlalonde:
Extract distributed_cache from discourse and add it to message_bus.


message_bus.gemspec
```diff
@@ -18,6 +18,7 @@ Gem::Specification.new do |gem|
   gem.version       = MessageBus::VERSION
   gem.required_ruby_version = ">= 2.2.0"
   gem.add_runtime_dependency 'rack', '>= 1.1.3'
+  gem.add_runtime_dependency 'concurrent-ruby'
```
jeremyevans: Like redis and pg, concurrent-ruby should be a development dependency and not a runtime dependency as it appears to be an optional feature.

nlalonde: Oh yeah, will do. Thanks.

PR #169, opened 2018-10-12
Add distributed_cache

nlalonde:
Extracted from discourse.

### DIFF
--- a/lib/message_bus/distributed_cache.rb
+++ b/lib/message_bus/distributed_cache.rb
@@ -1,0 +1,166 @@
+# frozen_string_literal: true
+
+# Like a hash, just does its best to stay in sync across the farm.
+# On boot all instances are blank, but they populate as various processes
+# fill it up.
+
+require 'weakref'
+require 'base64'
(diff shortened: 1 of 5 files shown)


lib/message_bus/distributed_cache.rb
```diff
@@ -0,0 +1,166 @@
...
+      site_id =
+        if @namespace
+          site_id_arg ||
+            (MessageBus.site_id_lookup && MessageBus.site_id_lookup.call) ||
+            DEFAULT_SITE_ID
+        else
+          DEFAULT_SITE_ID
+        end
+
+      @data[site_id] ||= Concurrent::Hash.new
```
SamSaffron: Interestingly we don't need this concurrent hash at all :) https://github.com/ruby-concurrency/concurrent-ruby/blob/master/lib/concurrent/hash.rb#L20 , so I guess just swap it with Hash, I guess technically for JRuby we may want to lock on get/set but we can hold off on that for now. nicer not to have the dependency anyway

PR #172, opened 2018-11-02
Workflow improvements

benlangfeld:


jeremyevans:
This is ultimately up to @SamSaffron, but personally I'm against checking in `Gemfile.lock`.  For libraries, any gems supported by `Gemfile` should be sufficient, there is no reason to lock specific versions, and it reduces the chance that people will test versions not in `Gemfile.lock`.

SamSaffron:
I am fine @jeremyevans  advice here, lets keep Gemfile.lock out 

phantom is a pretty sad state, I wonder if we can move to chrome driver

benlangfeld:
@jeremyevans Are you aware of the advice issued on the topic by the Bundler project? If so, do you still maintain that position?

The argument is that new versions of dependencies should be tested intentionally, and not accidentally by new contributors, who when checking out the project should have a known working configuration, rather than being test cases at a stage in their contribution which is most fragile and they are most likely to give up when they encounter a failing test suite.

@samsaffron Do you need the switch to Chrome in this PR?

benlangfeld:
I've removed the Gemfile.lock from this PR and will propose it in a separate PR if there is any movement on that discussion.

jeremyevans:
@benlangfeld I'm aware of the recommendations and do not agree with them.  It's certainly a tradeoff, but I'd rather new users hit failing cases and report those cases so they can be fixed, either in the code or by modifying the Gemfile/gemspec.  With a lock file, it's more likely such problems will never be found.

SamSaffron:
I would like to move to chrome driver asap, but for the time being I am good to merge this PR, thanks!

PR #174, opened 2018-11-07
Test improvements

benlangfeld:
Some of this was extracted from [the work on Redis Streams support](https://github.com/SamSaffron/message_bus/pull/173). I think it's worthwhile even before that happens.

Other parts generally improve the structure of the test suite (duplication of tests and conditional test definition rather than execution), it's communication of intent (a simpler forking spec) and its stability (avoiding regular failure in the forking spec on Redis).


spec/lib/message_bus/backend_spec.rb
```diff
@@ -212,19 +210,19 @@ def new_test_bus
     #   I thought about adding a subscribed callback, but outside of testing it matters less
     sleep 0.05
 
-    @bus.publish("/foo", "2")
+    @bus.publish("/foo", "two")
 
-    @bus.pub_redis.flushdb
+    @bus.reset!
```
SamSaffron: I much prefer reset methods, way cleaner

spec/lib/message_bus/backend_spec.rb
```diff
@@ -117,20 +117,57 @@ def new_test_bus
...
 
-    @bus.publish "/foo", "bar", max_backlog_age: 50
-    @bus.pub_redis.ttl(@bus.backlog_key("/foo")).must_be :<=, 50
-    @bus.pub_redis.ttl(@bus.backlog_key("/foo")).must_be :>, 0
+    # Start at time = 0s
+    @bus.publish "/foo", "bar", max_backlog_age: 1
+
+    @bus.global_backlog.length.must_equal 1
+
+    sleep 1.25 # Should now be at time =~ 1.25s. Our backlog should have expired by now.
```
SamSaffron: sleeps in tests make me uneasy, is there any other option here. even if we give a special "rush" method o back ends or something.

SamSaffron:
Overall I am fine with the changes it improves the test suite a lot 👍

spec/lib/message_bus/backend_spec.rb
benlangfeld: I'm not aware of a way to wind Redis' clock forward. The point of this is to actually exercise Redis' expiry rather than fake it in message_bus.

benlangfeld:
@SamSaffron If you're happy, I'm ready for this to be merged. I have another PR coming up to improve feature-parity.

SamSaffron:
Sure lets go for it! Thanks Ben

PR #175, opened 2018-11-07
Delint

benlangfeld:


lib/message_bus/rack/diagnostics.rb
```diff
@@ -32,23 +34,23 @@ def asset_path(asset)
...
-    #{js_asset "handlebars.js"}
-    #{js_asset "ember.js"}
-    #{js_asset "message-bus.js"}
-    #{js_asset "application.handlebars"}
-    #{js_asset "index.handlebars"}
-    #{js_asset "application.js"}
-  </body>
-</html>
-HTML
+    html = <<~HTML
```
SamSaffron: I am OK with dropping 2.2 support as it is EOL most of my gems have already done this, I know @jeremyevans  tends to keeps support for older version, I wonder if he is ok with this.

benlangfeld: We dropped support for 2.2 in https://github.com/SamSaffron/message_bus/commit/35a0cb093aef9b09317b24729da69ef97b514e99.

SamSaffron: forgot about that, we missed a spot https://github.com/SamSaffron/message_bus/commit/f07e4994270983cd08cc5d762be6ef5a41f40c67

benlangfeld:
This should be good to go :)

SamSaffron:
cool looks good. Mixed on the change it made to chat with the HTML being indented, but I guess consistency should win here.

PR #181, opened 2018-11-21
Fixes passing of custom pub sub options for publish

pawelniewie:
Previously didn't pass `max_backlog_size` when `max_backlog_age` was set.

Plus added site_id to another method.

### DIFF
--- a/lib/message_bus.rb
+++ b/lib/message_bus.rb
@@ -368,9 +368,10 @@ module MessageBus::Implementation
   # @yieldparam [MessageBus::Message] message each message as it is delivered
   #
   # @return [void]
-  def blocking_subscribe(channel = nil, &blk)
+  def blocking_subscribe(channel = nil, site_id = nil, &blk)
     if channel
-      reliable_pub_sub.subscribe(encode_channel_name(channel), &blk)
+      encoded_channel_name = encode_channel_name(channel, site_id)
+      reliable_pub_sub.subscribe(encoded_channel_name, &blk)
     else
       reliable_pub_sub.global_subscribe(&blk)
     end
--- a/lib/message_bus.rb
+++ b/lib/message_bus.rb
@@ -346,10 +346,10 @@ module MessageBus::Implementation
 
     channel_opts = nil
 
-    if opts && ((age = opts[:max_backlog_age]) || (size = opts[:max_backlog_size]))
+    if opts && (opts[:max_backlog_age] || opts[:max_backlog_size])
       channel_opts = {
-        max_backlog_size: size,
-        max_backlog_age: age
+        max_backlog_age: opts[:max_backlog_age],
+        max_backlog_size: opts[:max_backlog_size]
       }
     end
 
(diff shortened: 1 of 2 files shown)


SamSaffron:
The fix looks good, a bit concerned about the robocop change @benlangfeld are you ok with it? 

Also, I would very much prefer not to pull in the `spy` gem dependency, can you test this without it?

pawelniewie:
Rubocop was updated because I split one line into two following the same pattern as in https://github.com/SamSaffron/message_bus/blob/e57b2b2c972af844d6187203963e81ed638dfd35/lib/message_bus.rb#L258

You preferrer `rspec-mocks` or no new dependency at all? I thought having some mock/spy dep would be good for the future, but I understand if you don't want to add anything new there.

benlangfeld:
I’m happy with this. It would be nice if this was testable without a mock, though I’m not sure if it would be worth the much heavier test.

SamSaffron:
The change itself + robocop change is totally fine 👍  , but I would like `spy` removed and prefer no mocking library even if we carry a heavier integration test to check this.

spec/lib/message_bus_spec.rb
```diff
@@ -188,6 +188,22 @@
       site_id.must_equal 'law-and-order'
       channel.must_equal '/chuck'
     end
+
+    it "passes custom pub sub options" do
+      publish_args = nil
+
+      @bus.reliable_pub_sub.define_singleton_method :publish do |*args|
```
benlangfeld: @SamSaffron Are you happy with this style of stubbing?

SamSaffron: not really, I much prefer a full integration test here @pawelniewie hitting age and size should be fairly straight forward with the memory provider, the one thing I am perfectly comfortable mocking is time but we should have some examples doing that already imo.

lib/message_bus.rb
```diff
@@ -248,20 +248,21 @@ def publish(channel, data, opts = nil)
...
+        max_backlog_size: opts[:max_backlog_size]
       }
     end
 
     encoded_channel_name = encode_channel_name(channel, site_id)
     reliable_pub_sub.publish(encoded_channel_name, encoded_data, channel_opts)
   end
 
-  def blocking_subscribe(channel = nil, &blk)
+  def blocking_subscribe(channel = nil, site_id = nil, &blk)
```
benlangfeld: @pawelniewie Why this method in particular? There are several other methods which could have this change made, and indeed be tested as such. Would you mind splitting this out into a separate PR?

SamSaffron: also api wise I generally prefer `blocking_subscribe(channel = nil, site_id: nil, &blk)` these days but that is probably part of a larger discussion.

benlangfeld: @pawelniewie When you split that out, please also rebase on master and add appropriate documentation.

SamSaffron: @benlangfeld I think we should first try to get diagnostics / redis stream out, not against an overhaul but lets close off those things first. also probably worth tracking how much we would break on such an overhaul separately.  its a bit tricky as a pattern cause usually if there is only 1 param I really like unamed, and in some cases even N params can be perfect unamed ... eg: max_of(1,2,3,4)

PR #182, opened 2018-11-22
Consistent backend interface with documentation

benlangfeld:
Special attention when reviewing should be paid to the methods that were made private or removed on each backend implementation. When reviewing, taking each commit one at a time may be helpful.

### DIFF
--- a/lib/message_bus/backends/base.rb
+++ b/lib/message_bus/backends/base.rb
@@ -1,0 +1,207 @@
+# frozen_string_literal: true
+
+require "message_bus/backends"
+
+module MessageBus
+  module Backends
+    # Backends provide a consistent API over a variety of options for persisting
+    # published messages. The API they present is around the publication to and
+    # reading of messages from those backlogs in a manner consistent with
+    # message_bus' philosophy.
+    #
+    # The heart of the message bus, a backend acts as two things:
+    #
+    # 1. A channel multiplexer
+    # 2. Backlog storage per-multiplexed channel.
+    #
+    # Backends manage and expose multiple backlogs:
+    #
+    # * A backlog for each channel, in which messages that were published to
+    #   that channel are stored.
+    # * A global backlog, which conceptually stores all published messages,
+    #   regardless of the channel to which they were published.
+    #
+    # Backlog storage mechanisms and schemas are up to each individual backend
+    # implementation, and some backends store messages very differently than
+    # others. It is not necessary in order to be considered a valid backend,
+    # to, for example, store each channel backlog as a separate collection.
+    # As long as the API is presented per this documentation, the backend is
+    # free to make its own storage and performance optimisations.
+    #
+    # The concept of a per-channel backlog permits for lookups of messages in
+    # a manner that is optimised for the use case of a subscriber catching up
+    # from a message pointer, while a global backlog allows for optimising the
+    # case where another system subscribes to the firehose of messages, for
+    # example a message_bus server receiving all publications for delivery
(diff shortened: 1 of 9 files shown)


SamSaffron:
Do we need the giant indenting change? also the huge change in formatting makes it super hard to pick up on non indenting changes.

benlangfeld:
@SamSaffron The indenting change is all in the first commit. If you review the commits one by one, it's easy to follow what's going on. The indenting is in order to use the style of module/class declaration (nested) that the Ruby community understands as preferred.

lib/message_bus/backends/memory.rb
```diff
@@ -190,39 +190,36 @@ def initialize(config = {}, max_backlog_size = 1000)
...
+      # @param [String] channel the name of the channel to which the message should be published
+      # @param [JSON] data some data to publish to the channel. Must be an object that can be encoded as JSON
+      # @param [Hash] opts
+      # @option opts [Boolean] :queue_in_memory NOT SUPPORTED
+      # @option opts [Integer] :max_backlog_age (`self.max_backlog_age`) the longest amount of time a message may live in a backlog before beging removed, in seconds
+      # @option opts [Integer] :max_backlog_size (`self.max_backlog_size`) the largest permitted size (number of messages) for the channel backlog; beyond this capacity, old messages will be dropped
+      #
+      # @return [Integer] the channel-specific ID the message was given
       def publish(channel, data, opts = nil)
-        client = self.client
```
SamSaffron: this was there to erase method calls

lib/message_bus/backends/postgres.rb
```diff
@@ -237,41 +238,34 @@ def initialize(config = {}, max_backlog_size = 1000)
...
+      # @param [Hash] opts
+      # @option opts [Boolean] :queue_in_memory NOT SUPPORTED
+      # @option opts [Integer] :max_backlog_age (`self.max_backlog_age`) the longest amount of time a message may live in a backlog before beging removed, in seconds
+      # @option opts [Integer] :max_backlog_size (`self.max_backlog_size`) the largest permitted size (number of messages) for the channel backlog; beyond this capacity, old messages will be dropped
+      #
+      # @return [Integer] the channel-specific ID the message was given
       def publish(channel, data, opts = nil)
         # TODO in memory queue?
 
-        client = self.client
```
SamSaffron: same

lib/message_bus/backends/base.rb
```diff
@@ -0,0 +1,170 @@
...
+  module Backends
+    # The heart of the message bus, it acts as two things:
+    #
+    # 1. A channel multiplexer
+    # 2. Backlog storage per-multiplexed channel.
+    #
+    # Message ids are all sequencially increasing integers starting at 0.
+    #
+    # @abstract
+    class Base
```
SamSaffron: 👍  on having a new base interface

lib/message_bus/backends/memory.rb
benlangfeld: How so? I don't see how it's different.

SamSaffron:
👍  on all 3 of Jeremy's points. 

I also worry about exact duplicate documentation it will go out of sync, I think you can have links in documentation for yard so just use that . 

Once we get these points addressed I am happy to merge this in.

benlangfeld:
1. Done. I opted for raising a custom exception because otherwise I don't have a good way to test compliance with the interface.
2. Done (see https://github.com/lsegal/yard/issues/720 for details).
3. Done.

Will merge this then.

PR #185, opened 2018-11-28
Allows preventing MessageBus start-up

benlangfeld:
A re-run of https://github.com/SamSaffron/message_bus/pull/152.

This allows MessageBus to be included in an application which is not fully setup yet. Some use cases include:

* Manipulating databases (`rake db:drop` etc) when using the Postgres backend.
* Running rake tasks against partially built environments, such as when deploying to Kubernetes before a Redis container comes up.

TODO:
* [x] Usage documentation. Pending merge of https://github.com/SamSaffron/message_bus/pull/184

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -103,7 +103,7 @@ Metrics/MethodLength:
 # Offense count: 1
 # Configuration parameters: CountComments.
 Metrics/ModuleLength:
-  Max: 415
+  Max: 419
 
 # Offense count: 18
 Metrics/PerceivedComplexity:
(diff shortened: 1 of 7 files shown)


benlangfeld:
On testing this change in a minimal dummy application, I can't actually reproduce the scenario where the listener is started (or even the middleware initialized) in typical Rails rake tasks, which makes this change useless. I'm working on establishing what the difference is in my real-world app which causes it to be invoked, but for now I am closing this.

benlangfeld:
Ok, so it seems that in the `production` environment, differently than the `development` environment, Rails intializes the middleware stack during `rake db:migrate`.

PR #187, opened 2018-11-29
Keep a running changelog of changes on master since the last release

benlangfeld:
This puts the burden of writing change-logs on the contributor rather than the release manager, reducing the effort to cut a release and improving the accuracy of the changelog. It also means that changelog entries from now on are more likely to be useful in git blame when looking for the actual change that produced them.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,60 @@
+Unreleased
+
+- FEATURE: Much extra implementation documentation, and some improvements to usage documentation.
+- FEATURE: Improvements to development workflow:
+  - Fully docker-based development and testing, with no other dependencies.
+  - More stringent Rubocop enforcement and greater compliance.
+  - Testing supported Ruby versions
+  - Better test coverage of some features
+  - Improved testing and assertion of the differences between backends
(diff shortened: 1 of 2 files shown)


jeremyevans:
I agree with keeping a running CHANGELOG.  However, I don't think it should be modified by the contributor, it should be modified after merging by the maintainer doing the merging.  Otherwise you are likely to end up with merge conflicts for unrelated pull requests that are submitted close together.

SamSaffron:
Yes totally, maybe add a tiny note about this in the contributing section in the readme?

SamSaffron:
@jeremyevans  I think it heavily depends on the amount of PRs going, if the volume is not high I much prefer conflicts than forgetting to add something to the release notes

benlangfeld:
> However, I don't think it should be modified by the contributor, it should be modified after merging by the maintainer doing the merging.

This means git blame continues to point at a commit that has nothing to do with the one that introduced the change. I have found that being able to blame the changelog and immediately jump to the commit that introduced the change is very useful.

The PR traffic on this project is very low. I think conflicts will be somewhat rare and easily resolved.

PR #191, opened 2018-12-05
More lint rule enforcement

benlangfeld:
Changes to these entries in the TODO file are common in PRs.


SamSaffron:
hmmm robocop wise there has been a development: 

Technically I am only comfortable taking on this: https://github.com/discourse/prometheus_exporter/commit/5bc2e1e99e08063bc3c9bb364c77715bfea2d660 and changing lockstep with Discourse.

I totally get @jeremyevans concern I am though in a tiny bit of a different camp, we use https://github.com/prettier/prettier and I love love love never thinking about code formatting. I hate arguing about `"` vs `'` and honestly don't really care and prefer if a decision is made for me. 

My goal for discourse is to introduce either rufo or something else once it is ready for prime time and slowly introduce it into my various projects so I stop thinking about code formatting. 

I am huge unfan of the robocop "fixers" cause they are not going Syntax -> AST -> formatted code.

So let's change it so we inherit off Discourse for now and simply follow the rules we take for Discourse they are actually quite a bit laxer than what we have here

Issue #189, opened 2018-12-04
We should ship an optional HTTP client


SamSaffron:
@lowjoel built https://github.com/lowjoel/message_bus-client , @tgxworld is busy building something similar

I feel it makes much more sense just to ship an optional HTTP client with the repo, that way we can keep it working, even if we end up building two gems here from the single repo. 

`gem install message_bus-http_client`

```
require 'message_bus/http_client'
client = MessageBus::HttpClient.new('https://chat.samsaffron.com/')
client.subscribe('/test/channel') do |message|
   ...
end

client.unsubscribe('/test/channel')

callback = lambda{|m| p m}
client.subscribe('/channel', 77, &callback)
client.unsubscribe('/channel', &callback) 

```

Something like this... Bonus points for HTTP/2 ;p

ramontayag:
We're building something similar too, but it puts the heavy lifting on a background worker.

https://github.com/bloom-solutions/message_bus_client_worker

It's not complete yet (no long polling).

benlangfeld:
This is now on the master branch courtesy of https://github.com/SamSaffron/message_bus/pull/192. @SamSaffron It's going to be a couple of weeks before I have serious diagnostics improvements, so it may be worth releasing 2.2.0 with what we've got so far, and then a 2.3.0 in a few weeks time.

SamSaffron:
Yeah I think we should release a 2.2.0 @tgxworld maybe once you are around can you push a new version?

tgxworld:
2.2.0 has been released.

PR #192, opened 2018-12-06
FEATURE: Ruby MessageBus HTTP client.

tgxworld:
https://github.com/SamSaffron/message_bus/issues/189

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -39,21 +39,45 @@ task spec_client_js: 'jasmine:ci'
 backends = Dir["lib/message_bus/backends/*.rb"].map { |file| file.match(%r{backends/(?<backend>.*).rb})[:backend] } - ["base"]
 
 namespace :spec do
+  spec_files = Dir['spec/**/*_spec.rb']
+  integration_files = Dir['spec/integration/**/*_spec.rb']
+
   backends.each do |backend|
     desc "Run tests on the #{backend} backend"
     task backend do
       begin
         ENV['MESSAGE_BUS_BACKEND'] = backend
-        sh "#{FileUtils::RUBY} -e \"ARGV.each{|f| load f}\" #{Dir['spec/**/*_spec.rb'].to_a.join(' ')}"
+        sh "#{FileUtils::RUBY} -e \"ARGV.each{|f| load f}\" #{(spec_files - integration_files).to_a.join(' ')}"
       ensure
         ENV.delete('MESSAGE_BUS_BACKEND')
       end
     end
   end
+
+  desc "Run integration tests"
+  task :integration do
+    def port_available?(port)
+      server = TCPServer.open("0.0.0.0", port)
+      server.close
+      true
+    rescue Errno::EADDRINUSE
+      false
+    end
+
+    begin
+      ENV['MESSAGE_BUS_BACKEND'] = 'memory'
+      pid = spawn("bundle exec puma -p 9292 spec/fixtures/test/config.ru")
+      sleep 1 while port_available?(9292)
+      sh "#{FileUtils::RUBY} -e \"ARGV.each{|f| load f}\" #{integration_files.to_a.join(' ')}"
+    ensure
+      ENV.delete('MESSAGE_BUS_BACKEND')
+      Process.kill('TERM', pid) if pid
+    end
+  end
 end
 
 desc "Run tests on all backends, plus client JS tests"
-task spec: backends.map { |backend| "spec:#{backend}" } + [:spec_client_js]
+task spec: backends.map { |backend| "spec:#{backend}" } + [:spec_client_js, "spec:integration"]
 
 desc "Run performance benchmarks on all backends"
 task :performance do
--- a/lib/message_bus/http_client.rb
+++ b/lib/message_bus/http_client.rb
@@ -1,0 +1,337 @@
+require 'securerandom'
+require 'net/http'
+require 'json'
+require 'uri'
+require 'message_bus/http_client/channel'
+
+module MessageBus
+  # MessageBus client that enables subscription via long polling with support
+  # for chunked encoding. Falls back to normal polling if long polling is not
+  # available.
+  #
+  # @!attribute [r] channels
+  #   @return [Hash] a map of the channels that the client is subscribed to
+  # @!attribute [r] stats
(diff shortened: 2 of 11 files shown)


tgxworld:
Not sure if we want the client to be within a single repo or not but what I did here is to ship it as part of message_bus.

lib/message_bus/http_client.rb
```diff
@@ -0,0 +1,268 @@
...
+  #
+  # @return [Integer] the current status of the client
+  def stop
+    @status = STOPPED
+  end
+
+  # Subscribes to a channel which executes the given callback when a message
+  # is published to the channel
+  #
+  # Example:
```
benlangfeld: Can you use `@example`?

lib/message_bus/http_client.rb
```diff
@@ -0,0 +1,268 @@
...
+  #   client.subscribe("/test", last_message_id: -2) do |payload|
+  #     puts payload
+  #   end
+  #
+  # @param channel [String] channel to listen for messages on
+  # @param last_message_id [Integer] last message id to start polling on.
+  # @yield [data, global_id, message_id]
+  #  callback to be executed whenever a message is received
```
benlangfeld: You should document the `@yieldparam`s also.

lib/message_bus/http_client/channel.rb
```diff
@@ -0,0 +1,12 @@
+module MessageBus
+  class HTTPClient
```
benlangfeld: We should document these, at a minimum to mark them as `@private`.

Rakefile
```diff
@@ -60,3 +59,23 @@ end
...
+  def port_available?(port)
+    server = TCPServer.open("0.0.0.0", port)
+    server.close
+    true
+  rescue Errno::EADDRINUSE
+    false
+  end
+
+  begin
+    ENV['MESSAGE_BUS_BACKEND'] = 'memory'
```
benlangfeld: Is it possible to also run these against other backends?

tgxworld: It is possible but I don't think we need to do so though because I'm only interested in testing the HTTP client. The backend is configured here https://github.com/SamSaffron/message_bus/pull/192/files#diff-926cc5ae85da76b1b9c6ba536922080eR3.

tgxworld:
@benlangfeld  Thank you so much for reviewing :+1: I've updated the PR as per your comments.

Rakefile
```diff
@@ -39,17 +38,41 @@ task spec_client_js: 'jasmine:ci'
...
-        sh "#{FileUtils::RUBY} -e \"ARGV.each{|f| load f}\" #{Dir['spec/**/*_spec.rb'].to_a.join(' ')}"
+        sh "#{FileUtils::RUBY} -e \"ARGV.each{|f| load f}\" #{(spec_files - integration_files).to_a.join(' ')}"
       ensure
         ENV.delete('MESSAGE_BUS_BACKEND')
       end
     end
   end
+
+  desc "Run integration tests"
+  task :integration do
```
benlangfeld: Surely we should run these tests in CI?

Rakefile
```diff
@@ -1,7 +1,6 @@
 require 'rubygems'
 require 'rake/testtask'
 require 'bundler'
-require 'bundler/gem_tasks'
```
benlangfeld: Is there some other release process besides these?

lib/message_bus/http_client.rb
```diff
@@ -0,0 +1,297 @@
...
+    STATUS_CHANNEL = "/__status".freeze
+    private_constant :STATUS_CHANNEL
+
+    STOPPED = 0
+    STARTED = 1
+
+    Stats = Struct.new(:failed, :success)
+    private_constant :Stats
+
+    # @param base_url [String] Base URL of the message_bus server to connect to
```
benlangfeld: What are the other args?

lib/message_bus/http_client/configuration.rb
```diff
@@ -0,0 +1,40 @@
+module MessageBus
+  class HTTPClient
+    module Configuration
```
benlangfeld: What is the benefit of this being a separate module if it is only used once?

tgxworld: It makes the main file easier to read and I avoid having to flood the main class with more comments for yard.

benlangfeld: But all of these options are important to know about when instantiating the class. Just seems like unnecessary indirection.

Rakefile
tgxworld: Oops good catch. I

tgxworld:
@benlangfeld  Applied the changes per your review. Thank you for taking the time to review this PR :+1:

lib/message_bus/http_client.rb
```diff
@@ -0,0 +1,324 @@
...
+  #   @return [Stats] a Struct containing the statistics of failed and successful
+  #     polling requests
+  class HTTPClient
+    class InvalidChannel < StandardError; end
+    class MissingBlock < StandardError; end
+
+    attr_reader :channels,
+                :stats
+
+    attr_accessor :enable_long_polling,
```
benlangfeld: Can we document these or make them private?

tgxworld: Oops don't have the habit of writing docs for yard so I missed this out. Fixed!

SamSaffron:
@jeremyevans FYI we are going to have to do a breaking release here for you cause you will need to explicitly require `message_bus/middleware` from now on. (rails do not need to cause of the railstie)

I feel this is the right thing to do cause autoload is going away and this opens the ability of using message bus without HTTP at all. 

Do you think its OK to release 2.2 here or do we need to bump all the way up to 3.0? 

No APIs broke BUT blanket upgrade will lead to errors for rack app cause they will need to require the middleware.

SamSaffron:
Note Jeremy this comment was about the stripping of the explicit rack dependency from message bus which was in another PR. But shipping with the optional HTTP client is the underlying reason for the change.

jeremyevans:
I would definitely go to 3.0 if we expect a good portion of apps would break after the change.  It looks like `roda-message_bus` already requires `message_bus/rack/middleware`, so I'm not sure if it will need any changes.

PR #190, opened 2018-12-05
Performance tests

benlangfeld:
Executing this on my laptop I get this:

```
docker-compose run tests rake performance
Starting message_bus_postgres_1 ... done
Starting message_bus_redis_1    ... done
/usr/local/bin/ruby -e "ARGV.each{|f| load f}" spec/performance/publish.rb
Running publication benchmark with 10000 iterations on backends: [:postgres, :redis, :memory]

                                                                   user     system      total        real
postgres - publication only                                    2.880000   1.640000   4.520000 ( 26.512851)
redis - publication only                                       0.870000   0.690000   1.560000 (  3.890986)
memory - publication only                                      1.130000   0.000000   1.130000 (  1.142721)

postgres - subscription no trimming                            3.000000   1.970000   4.970000 ( 28.369772)
redis - subscription no trimming                               2.510000   1.180000   3.690000 (  6.072367)
memory - subscription no trimming                              2.900000   0.000000   2.900000 (  2.909922)

postgres - subscription with trimming                          3.620000   2.030000   5.650000 ( 28.796871)
redis - subscription with trimming                             2.050000   0.940000   2.990000 (  5.231705)
memory - subscription with trimming                            1.340000   0.000000   1.340000 (  1.346209)

[postgres - subscription no trimming]: 10000 messages sent, 10000 received, rate of 100.0%
[redis - subscription no trimming]: 10000 messages sent, 10000 received, rate of 100.0%
[memory - subscription no trimming]: 10000 messages sent, 10000 received, rate of 100.0%

[postgres - subscription with trimming]: 10000 messages sent, 10000 received, rate of 100.0%
[redis - subscription with trimming]: 10000 messages sent, 10000 received, rate of 100.0%
[memory - subscription with trimming]: 10000 messages sent, 10000 received, rate of 100.0%
```

PR #193, opened 2018-12-06
Executes performance tests in CI

benlangfeld:
I'd like to revisit this in isolation. It was unclear in https://github.com/SamSaffron/message_bus/pull/190 if the objection was to execution in the default task (which perhaps people run when working on message_bus) or execution in CI.

CI is the quietest most stable environment we have available in which to execute these. Development hardware is noisy, and we have no other environment readily available.

Sam mentioned the possibility of using https://rubybench.org/, but from that site I found no evidence that it is an option at this time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ env:
   - PGUSER=postgres
 before_script:
   - psql -c 'create database message_bus_test;' -U postgres
+script: rake ci
 services:
   - redis-server
   - postgresql
--- a/Rakefile
+++ b/Rakefile
@@ -67,3 +67,5 @@ end
 
 desc "Run all tests, link checks and confirms documentation compiles without error"
 task default: [:spec, :rubocop, :test_doc]
+
+task ci: [:default, :performance]


benlangfeld:
@jeremyevans How does one sign up to include these results on RubyBench? From what I can see, it's not available. Do you have a suggestion for anywhere other than Travis that these could run?

jeremyevans:
Not sure in regards to getting included in RubyBench.  FWIW, even if no online service could run the benchmarks, I still don't think they should be run in CI.  IMO, CI is for automatically reporting test failures.  Unless the benchmarks were rewritten as performance tests, where performance below a certain threshold results in failure and will be reported, I don't think they are appropriate for CI.  Considering the virtualized environment running the CI, CI performance is probably too variable to write robust performance tests.

SamSaffron:
I feel like CI is fundamentally the wrong spot to run this. You have no guarantee what AWS instance you will get, how noisy the neighbors are and so on. A 40% slowdown can easily be you just happened to hit the wrong instance 

Instead... @tgxworld  can help guide you at what we need to do to add this to ruby bench, he maintains the project :)

Issue #186, opened 2018-11-29
Diagnostics improvements


benlangfeld:
I'm planning on implementing some improvements to diagnostics to expand it into more of a dashboard similar to the Resque one, providing the ability to browse backlogs, see stats on backlog length, trimming/expiry, recently connected clients, etc.

@SamSaffron Please let me know if you have any thoughts on this as I get started.

The first thing I ran into is that the versions of Ember and Handlebars currently in use are from 2013 and are missing useful features. Upgrading them may be non-trivial, and I am more accustomed to the React landscape myself. Would you be against switching the diagnostics functionality to the latest version of React, @SamSaffron ?

SamSaffron:
I am fine switching to react here, whatever you prefer. You would be pretty much rebooting this project anyway.

PR #194, opened 2018-12-06
Re-write diagnostics in React

benlangfeld:
This is the beginning of https://github.com/SamSaffron/message_bus/issues/186. Submitting it right away just in case there's some fundamental issue before I really get stuck in.


SamSaffron:
Nothing is fundamentally wrong here 👍  only suggestion I have is can you include screenshots as you go, it makes it easier to provide quick feedback

benlangfeld:
I will include screenshots as I work. Right now, this has feature parity with the Ember version and looks the same.

PR #197, opened 2018-12-10
Remove gem's dependency on rack.

tgxworld:


tgxworld:
Spoke with @SamSaffron about this and since this is a breaking change, we'll need to do a major version bump. However, we're currently in `2.2.0.pre` so we'll have to wait for that version to land first.

PR #199, opened 2019-02-19
add typings

yoyomo:
a Typescript file for typescript users

### DIFF
--- a/typings/message-bus.d.ts
+++ b/typings/message-bus.d.ts
@@ -1,0 +1,37 @@
+declare interface MessageBus {
+  start(): void
+
+  subscribe(channel: string, func: (payload?: any | void, globalId?: string | void, messageId?: string | void) => void, lastId?: number | void): void
+
+  unsubscribe(channel?: string, func?: () => void): void
+
+  pause(): void
+
+  resume(): void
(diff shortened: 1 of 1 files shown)


typings/message-bus.d.ts
```diff
@@ -0,0 +1,37 @@
+declare interface MessageBus {
```
benlangfeld: How do we keep this in sync with both the implementation and the documentation? Is there tooling for checking that, or some way we can work it into our tests? Alternatively, is it possible to include this inline with the existing implementation and parse it out, such that it’s easier to spot when method signatures change but the type definition does not?

benlangfeld: If this were merged, how would someone use it?

SamSaffron:
I agree with Ben's concerns here, I think the most compelling argument here is just to have the original message bus in typescript and during build compile it to js. Then allow people to either import .ts or .js. I am not categorically against this but it is an involved change and we need to prove that the js generated is exactly the same as the current js file.

ZogStriP:
Closing this. @yoyomo Feel free to re-open another PR if you port message bus to TS.

PR #204, opened 2019-04-17
update README 

tgodfrey:
Pointer to a gist with an approach for publishing to Redis from Elixir for MessageBus to pick up the message.


SamSaffron:
awesome thanks !

PR #205, opened 2019-04-26
on_middleware_error only for internal errors

benlangfeld:
Prior to this change, the `on_middleware_error` error handler was being invoked on all exceptions that hit this middleware, including those which come from lower-level middleware. This has the side-effect of potentially masking exceptions from the Rails application, and is surprising because `on_middleware_error` is assumed to handle only message_bus errors.


SamSaffron:
I am a bit confused, what changed here?

benlangfeld:
The first line was removed from the rescue clause.

SamSaffron:
I yeah, just read through it carefully, yes nice fix, probably worth adding a test for this!

benlangfeld:
@SamSaffron I committed tests to master. Would you mind cutting a release?

SamSaffron:
Sure I just did! Thanks

Issue #208, opened 2019-06-04
Use of `Redis#client` method which was removed in redis v4.0


chrismhilton:
We're using `message_bus` with Redis as the storage backend and have recently updated our `redis` gem from v3.3.3 to v4.1.0

In v4.0 of the `redis` gem the `Redis#client` method was replaced with `Redis#_client` which meant it was also necessary for us to update our `message_bus` gem from v2.0.2 to v2.1.6 as in the older version the `after_fork` method used `pub_redis.client.reconnect`

We're all good now but looking through the gem code I did notice that the `is_readonly?` method is still using `pub_redis.client`

```
def is_readonly?
  key = "__mb_is_readonly"

  begin
    # in case we are not connected to the correct server
    # which can happen when sharing ips
    pub_redis.client.reconnect
    pub_redis.client.call([:set, key, '1'])
    false
  rescue Redis::CommandError => e
    return true if e.message =~ /^READONLY/
  end
end
```

This means that with `redis` v4.0 onwards, when the method calls `pub_redis.client.reconnect` it will result in a `Redis::CommandError` exception with the message `ERR Unknown subcommand or wrong number of arguments for ''. Try CLIENT HELP` and the method will always return `nil`.

SamSaffron:
Nice catch! Can you send a pr through to fix?

PR #209, opened 2019-06-05
Fix for the redis backend `is_readonly?` method

chrismhilton:
Fix to make the `is_readonly?` method compatible with the `redis` gem both pre and post v4.0 when the `client` attribute was removed

This resolves issue #208

For further info this is the commit in the `redis` gem that removed the `client` attribute:
https://github.com/redis/redis-rb/commit/c239abb43c2dce99468bf94652a3c48b31a1041a


SamSaffron:
Thanks, this looks good to me!

PR #214, opened 2019-10-18
FIX: Protect Redis config from being manipulated

lis2:
Redis gem is allowing to pass custom connector as an option. Later, code is removing that option and initialize custom connector:
```
options.delete(:connector).new(@options) # https://github.com/redis/redis-rb/blob/master/lib/redis/client.rb#L95
```

It works for MessageBus when Redis server is running. When it fails and MessageBus is trying to reconnect, because it is a passing reference to Redis, customer connector is not available anymore.

Therefore, I think it would be better to pass a copy of the original object.

### DIFF
--- a/lib/message_bus/backends/redis.rb
+++ b/lib/message_bus/backends/redis.rb
@@ -327,7 +327,7 @@ LUA
       private
 
       def new_redis_connection
-        ::Redis.new(@redis_config)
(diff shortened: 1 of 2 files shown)


SamSaffron:
This looks good to me, odd that redis are fiddling like this with config, seems unsafe.

PR #215, opened 2019-10-20
Deprecated methods and rubocop

lis2:
Housekeeping, deprecated Minitest methods and Rubocop magic comment

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,6 +7,7 @@ gemspec
 group :test do
   gem 'minitest'
   gem 'minitest-hooks'
+  gem 'minitest-global_expectations'
   gem 'rake'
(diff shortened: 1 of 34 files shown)


jeremyevans:
Alternatively, if we don't want to uglify the specs, we could use `minitest-global_expectations` and we would only need to change a couple lines.

SamSaffron:
I am with Jeremy here @lis2, lets just pull in mini-test-global_expectations 

This new pattern does make the spec suite quite ugly.

lis2:
@jeremyevans @SamSaffron fair point :) fixed

SamSaffron:
Looks great now! thanks!

PR #216, opened 2020-01-01
Allow access to http_client thread so it can be waited on

wishdev:


benlangfeld:
What is the use case?

wishdev:
> What is the use case?

Something as simple as a client that simply wants to gather messages

client = MessageBus::HTTPClient.new("http://a.ca")
client.subscribe('/bob") { do_something }

The next line is the issue - the client is running on a background thread so I need to do something to keep the app alive. I can sleep, loop, or with this PR simply thread#join on the client thread.

The StopError exception also lets the client properly exit the thread as well if a message to exit came through. Something like

client.subscribe("/exit") { client.stop } would immediately exit the thread via the exception and shut down the connection. I'm sure I could just exit the app but if the app wanted a cleaner way to shut things down (i.e. it does other work after getting the exit message) I think the exception provides a method to accomplish that.

SamSaffron:
Do we want to simplify the API a bit to avoid leaking out implementation details? 

Maybe simply allow `client.wait` which can call the Thread.join.

Then if our implementation ever changes we don't need to be tied to this very specific detail?

SamSaffron:
Sorry, closing this as I don't want to leak implementation details out of the class, thread is a detail. Instead a PR to add .wait sound A-OK to me and that can do the thread.join .

PR #217, opened 2020-02-13
README.md: Doc adding extra response headers

henrik:
Took us a while to find this one, so we thought mentioning it in the README could be helpful.


henrik:
Another thing that might help would be a link to https://www.rubydoc.info/gems/message_bus. The code seems well documented but a link saves generating docs locally.

SamSaffron:
great example, yes please! any PRs to improve documentation and discovery of documentation is mega welcome!

PR #219, opened 2020-04-24
FIX: Don't publish message to intersection of `user_ids` and `group_ids`

tgxworld:
The original design intent was to always publish messages to clients
that are either in `user_ids` and `group_ids`. However, there is a long
standing bug where we only published to the intersection of `user_ids`
and `group_ids` which this commit seeks to resolve.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,38 +74,62 @@ id = MessageBus.last_id("/channel")
 MessageBus.backlog "/channel", id
 ```
 
+### Targetted messages
+
+Messages can be targetted to particular clients by supplying the `client_ids` option when publishing a message.
+
 ```ruby
-# messages can be targetted at particular users or groups
-MessageBus.publish "/channel", "hello", user_ids: [1,2,3], group_ids: [4,5,6]
+MessageBus.publish "/channel", "hello", client_ids: ["XXX", "YYY"] # (using MessageBus.clientId)
+```
 
-# messages can be targetted at particular clients (using MessageBus.clientId)
-MessageBus.publish "/channel", "hello", client_ids: ["XXX","YYY"]
+By configuring the `user_id_lookup` and `group_ids_lookup` options with a Proc or Lambda which will be called with a [Rack specification environment](https://github.com/rack/rack/blob/master/SPEC.rdoc#the-environment-), messages can be targetted to particular clients users or groups by supplying either the `user_ids` or `group_ids` options when publishing a message.
 
-# message bus determines the user ids and groups based on env
-
+```ruby
 MessageBus.configure(user_id_lookup: proc do |env|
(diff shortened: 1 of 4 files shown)


lib/message_bus/client.rb
```diff
@@ -128,15 +128,25 @@ def <<(msg)
...
+    elsif allow_all_group_ids
+      allow_all_user_ids || msg.user_ids.include?(self.user_id)
+    else
       (
-        msg.group_ids - self.group_ids
-      ).length < msg.group_ids.length
-    )
+        (msg.group_ids - self.group_ids).length < msg.group_ids.length
+      ) || msg.user_ids.include?(self.user_id)
+    end
```
SamSaffron, line 149: This logic is a bit hard to reason about ... maybe:

```
  def allowed?(msg)
    client_allowed = !msg.client_ids || msg.client_ids.include?(self.client_id)

    user_allowed = false
    group_allowed = false

    # this is an inconsistency we should fix anyway, publishing `user_ids: nil` should work same as groups
    has_users = msg.user_ids && msg.user_ids.length > 0
    has_groups = msg.group_ids && msg.group_ids.length > 0

    if has_users
      user_allowed = msg.user_ids.include?(self.user_id)
    end

    if has_groups
      group_allowed = (
        msg.group_ids - self.group_ids
      ).length < msg.group_ids.length
    end

    client_allowed && (user_allowed || group_allowed || (!has_users && !has_groups))
  end

```

.travis.yml
```diff
@@ -1,7 +1,6 @@
 before_install: gem install bundler
 language: ruby
 rvm:
-  - 2.3
```
tgxworld, line 4: Our latest Rubocop rule is not compatible with Ruby 2.3 so I'm dropping support since support for Ruby 2.3 has already ended.

README.md
```diff
@@ -74,39 +74,63 @@ id = MessageBus.last_id("/channel")
...
 end
 ```
 
+If both `user_ids` and `group_ids` options are supplied when publishing a message, the message will be targetted at clients with lookup return values that  matches on either the `user_ids` **or** the `group_ids` options.
+
+```ruby
+MessageBus.publish "/channel", "hello", user_ids: [1, 2, 3], group_ids: [1, 2, 3]
+```
+
+If the `client_ids` option is supplied with either the `user_ids` or `group_ids` options when publising a message, the `client_ids` option will take precendence over both options.
```
SamSaffron, line 124:
```suggestion
If the `client_ids` option is supplied with either the `user_ids` or `group_ids` options when publising a message, the `client_ids` option will be be applied unconditionally and messages will be filtered further using `user_id` or `group_id` clauses. 
```

PR #220, opened 2020-04-28
FEATURE: `MessageBus#register_message_filter`

tgxworld:


README.md
```diff
@@ -129,6 +129,20 @@ MessageBus.publish "/channel", "hello", client_ids: ["XXX", "YYY"], user_ids: [1
...
 Passing `nil` or `[]` to either `client_ids`, `user_ids` or `group_ids` is equivalent to allowing all values on each option.
 
+### Filtering Messages
+
+Custom message filters can be registered via `MessageBus#register_message_filter`. This can be useful for filtering away messages from the client based on the message's payload.
+
+For example we can ensure that only messages seen by the server in the last 20 seconds are published to the client:
+
+```
+MessageBus.register_message_filter(/^\/test\/\d+$/) do |message|
```
SamSaffron, line 139: let's call it `register_client_message_filter` ... since it only applies to clients and does not apply to stuff like just calling backlog or subscribe.

lib/message_bus.rb
```diff
@@ -554,6 +554,25 @@ def keepalive_interval
...
   end
 
+  # Registers a message filter that allows messages to be filtered from the client.
+  #
+  # @param [Regexp] a regular expression to match against a message's channel
+  #
+  # @yieldparam [MessageBus::Message] message published to the channel that matched the regular expression
+  # @yieldreturn [Boolean] whether the message should be published to the client
+  # @return [void]
+  def register_message_filter(regexp, &blk)
```
SamSaffron, line 564: maybe we don't need a regex ... start_with with a string should be good enough and a touch faster anyway.

`register_client_message_filter(channel_prefix, &blk)`

lib/message_bus.rb
```diff
@@ -554,6 +554,25 @@ def keepalive_interval
...
   end
 
+  # Registers a message filter that allows messages to be filtered from the client.
+  #
+  # @param [Regexp] channel_regexp regular expression to match against a message's channel
+  #
+  # @yieldparam [MessageBus::Message] message published to the channel that matched the regular expression
+  # @yieldreturn [Boolean] whether the message should be published to the client
+  # @return [void]
+  def register_message_filter(channel_regexp, &blk)
```
SamSaffron, line 564: `register_client_message_filter(channel_prefix, &blk)` no need to support full regex here we can just use string.start_with? which is faster.

lib/message_bus.rb
```diff
@@ -554,6 +554,25 @@ def keepalive_interval
     @config[:keepalive_interval] || 60
   end
 
+  # Registers a client message filter that allows messages to be filtered from the client.
+  #
+  # @param [String] channel_prefix channel prefix to match against a message's channel
```
SamSaffron, line 559: technically we are allowing String or Regexp

tgxworld: O I just ran into this but if we support Ruby 2.4, we can't pass it a Regexp

SamSaffron: no probs

SamSaffron:
looks good

PR #222, opened 2020-05-12
Allow the base route to be altered

### LINKED ISSUE #202: Unable to configure and use baseurl in rails application
Hello,
I have configured baseurl in my application js file like.
 `
   MessageBus.baseUrl = 'fa/';
    MessageBus.start();
    MessageBus.subscribe("/notification_channel_" + window.wsConfig.sessionId, function(data){
        console.log(data);
    }
`
I keep getting following error in my Rails application
`ActionController::RoutingError` (No route matches [POST] `"/fa/message-bus/fc1684579dd54b9bb0a89fd8e4a3fa5d/poll.`

I tried exploring lib/message_bus/rack/middleware.rb file but could not find any code handling response with custom baseurl. Please help. Thanks

nikolai-b:
Resolves #202

Allows the route that message bus listens on to be configured

### DIFF
--- a/lib/message_bus/rack/middleware.rb
+++ b/lib/message_bus/rack/middleware.rb
@@ -55,7 +55,7 @@ class MessageBus::Rack::Middleware
   # Process an HTTP request from a subscriber client
   # @param [Rack::Request::Env] env the request environment
   def call(env)
-    return @app.call(env) unless env['PATH_INFO'] =~ /^#{@bus.base_route}message-bus\//
+    return @app.call(env) unless env['PATH_INFO'].start_with? "#{@bus.base_route}message-bus/"
 
     handle_request(env)
   end
--- a/lib/message_bus/rack/middleware.rb
+++ b/lib/message_bus/rack/middleware.rb
@@ -39,7 +39,7 @@ class MessageBus::Rack::Middleware
     @bus = config[:message_bus] || MessageBus
     @connection_manager = MessageBus::ConnectionManager.new(@bus)
     @started_listener = false
-    @client_id_regexp = Regexp.new(@bus.base_route + 'message-bus/(?<client_id>[[:alnum:]]+)')
+    @base_route_length = "#{@bus.base_route}message-bus/".length
     start_listener unless @bus.off?
   end
 
@@ -75,7 +75,7 @@ class MessageBus::Rack::Middleware
       return diags.call(env)
     end
 
-    client_id = env['PATH_INFO'].match(@client_id_regexp) { |m| m[:client_id] }
+    client_id = env['PATH_INFO'][@base_route_length..-1].split("/")[0]
     return [404, {}, ["not found"]] unless client_id
 
     user_id = @bus.user_id_lookup.call(env) if @bus.user_id_lookup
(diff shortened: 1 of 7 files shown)


SamSaffron:
I think this looks ok, can you add a note to release notes and readme about this?

lib/message_bus/rack/middleware.rb
```diff
@@ -54,7 +54,7 @@ def stop_listener
   # Process an HTTP request from a subscriber client
   # @param [Rack::Request::Env] env the request environment
   def call(env)
-    return @app.call(env) unless env['PATH_INFO'] =~ /^\/message-bus\//
+    return @app.call(env) unless env['PATH_INFO'] =~ /^#{@bus.base_route}message-bus\//
```
SamSaffron, line 57: might as well memoize this regex

SamSaffron: I wonder do we even need a regex here ... why not just use start_with?

nikolai-b:
Not sure why CI is failing, looks like it got stuck

lib/message_bus/rack/middleware.rb
```diff
@@ -39,6 +39,7 @@ def initialize(app, config = {})
     @bus = config[:message_bus] || MessageBus
     @connection_manager = MessageBus::ConnectionManager.new(@bus)
     @started_listener = false
+    @client_id_regexp = Regexp.new(@bus.base_route + 'message-bus/(?<client_id>[[:alnum:]]+)')
```
SamSaffron, line 42: I think you need a regex escape here ... base route could be `test.it` ... then this would match `test7it`

nikolai-b: Ah, thanks.  I've changed it to using a substring as I was also thinking that client_id might have `-` or other characters.  I thought it was slower but looks like it is about the same (or faster even)

```ruby
base_route = "/base/route/";
path_env = "#{base_route}message-bus/clientId";

rgx = Regexp.new(base_route + 'message-bus/(?<client_id>[[:alnum:]]+)');

substr_proc = -> { path_env["#{base_route}message-bus/".length..-1].split("/")[0] };
regex_proc = -> { path_env.match(rgx) { |m| m[:client_id] } };
split_proc = -> { path_env.split("#{base_route}message-bus/")[1].split("/")[0] };

substr_proc.call # gives "clientId"
regex_proc.call # gives "clientId"
split_proc.call # gives "clientId"

Benchmark.ips { |x| x.report("substr") { substr_proc.call }; x.report("regexp") { regex_proc.call }; x.report("split") { split_proc.call } };
```

```
Warming up --------------------------------------
              substr    75.167k i/100ms
              regexp    67.942k i/100ms
               split    65.618k i/100ms
Calculating -------------------------------------
              substr      1.004M (± 8.0%) i/s -      5.036M in   5.048246s
              regexp    801.254k (±13.0%) i/s -      3.941M in   5.013270s
               split    799.437k (± 6.1%) i/s -      4.003M in   5.025920s
```

lib/message_bus/rack/middleware.rb
```diff
@@ -54,7 +55,7 @@ def stop_listener
   # Process an HTTP request from a subscriber client
   # @param [Rack::Request::Env] env the request environment
   def call(env)
-    return @app.call(env) unless env['PATH_INFO'] =~ /^\/message-bus\//
+    return @app.call(env) unless env['PATH_INFO'].start_with? "#{@bus.base_route}message-bus/"
```
SamSaffron, line 58: I know I am being ultra stickler here, but might as well micro optimise this cause it is called on every request.

Maybe memoize this value at `@bus.base_path` then you just keep one of these strings around for the life of the process.

SamSaffron:
OK lets give this a shot! thanks @nikolai-b !

PR #225, opened 2020-05-25
Clean up Javascript

eviltrout:
This PR is mostly about cleaning up `var`s, global state and a broken function from the front end JS. It's a step towards making this a true Javascript module.


SamSaffron:
This all looks really good to me!

PR #226, opened 2020-05-26
FEATURE: Support UMD and NPM

eviltrout:
This gives us a package.json we can publish to NPM as well as a
universal module definition that supports browser globals as well as
ES6 imports in Ember CLI via auto import.

### DIFF
--- a/assets/message-bus.js
+++ b/assets/message-bus.js
@@ -1,5 +1,19 @@
 /*jshint bitwise: false*/
-(function(exports, document, jQuery) {
+
+(function (root, factory) {
+  if (typeof define === 'function' && define.amd) {
+    // AMD. Register as an anonymous module.
+    define([], function (b) {
+      // Also create a global in case some scripts
+      // that are loaded still are looking for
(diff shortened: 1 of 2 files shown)


SamSaffron:
Looks good, merge at will, maybe add a rake task to publish to NPM?

eviltrout:
@SamSaffron I didn't think that was necessary because npm is already a command line task - it's `npm publish`. Would you prefer a rake task that runs that simple command?

SamSaffron:
I think ... we should inject into `bundle exec rake:release` just so we don't forget when we do new gem releases. 

The tricky thing though is that most gem releases do not touch the js bundle... but some do.

There is an advantage though of having version parity, so you know for sure that message bus 1.2.1 works well with npm message bus 1.2.1 ... eliminates a bunch of thinking.

eviltrout:
Added here:

https://github.com/discourse/message_bus/commit/7847936a892eddde0fae1b7d2e28a0bae0452a21

PR #229, opened 2020-06-05
FIX: MessageBus backend can leak Redis connection.

tgxworld:


lib/message_bus/backends/redis.rb
```diff
@@ -253,9 +253,15 @@ def subscribe(channel, last_id = nil)
...
         if @redis_global
           # new connection to avoid deadlock
-          new_redis_connection.publish(redis_channel_name, UNSUB_MESSAGE)
-          @redis_global.disconnect
-          @redis_global = nil
+          begin
+            new_redis = new_redis_connection
+            new_redis.publish(redis_channel_name, UNSUB_MESSAGE)
+            @redis_global.disconnect
+            @redis_global.disconnect!
```
SamSaffron, line 260: this double disconnect is confusing, why do we need it?

lib/message_bus/backends/redis.rb
```diff
@@ -278,7 +284,7 @@ def global_subscribe(last_id = nil, &blk)
         end
 
         begin
-          @redis_global = new_redis_connection
+          @redis_global ||= new_redis_connection
```
SamSaffron, line 287: hmmm this though is changing behavior, if there are 2 global subscribes then they share a connection so this will get stuck. 

I guess instead if there is a redis global already we could either exception out or disconnect it, or hold an array of redis globals

SamSaffron:
overall very good catch here!

lib/message_bus/backends/redis.rb
tgxworld: `@redis_global.disconnect` disconnects from the Redis sub. While `@redis_global.disconnect!` actually disconnects the client.

SamSaffron: won't disconnect! do both?

tgxworld: That is a good point. I guess we just want to close the connection here.

tgxworld: Actually we can't, we need to `disconnect` from the sub first because calling `disconnect!` will cause an error to be thrown.

SamSaffron: yikes, that feels like a redis gem bug to me... disconnect! is meant to be rude and error less. Maybe do a PR to redis gem and then link to it from the source?

tgxworld: O actually `@redis_global.disconnect` is meant to wait until we unsubscribe from the subscription here. The error thrown if we straight up `disconnect!` is because our `on.unsubscribe` callback which calls `@redis_global.unsubscribe` will be working with a `nil` object.

SamSaffron: maybe we fix that? then we don't need a double disconnect?

lib/message_bus/backends/redis.rb
tgxworld: It isn't possible to have 2 global subscribers from what I can tell. Each instance of messageBus will only have one subscriber thread.

```
  def ensure_subscriber_thread
    @mutex.synchronize do
      return if (@subscriber_thread && @subscriber_thread.alive?) || @destroyed

      @subscriber_thread = new_subscriber_thread
    end
  end
```

SamSaffron: yeah so `||=` feels odd here?

tgxworld: OOOOO but we allow people to do a blocking subscribe...

```
  def blocking_subscribe(channel = nil, &blk)
    if channel
      reliable_pub_sub.subscribe(encode_channel_name(channel), &blk)
    else
      reliable_pub_sub.global_subscribe(&blk)
    end
  end
```

tgxworld: I need to do a full review of the code path here. With `blocking_subscribe`, the `@redis_global` can be reassigned and doesn't seem to be thread safe.

SamSaffron: yikes, I guess to be 100% correct we need to start tracking an array of `redis_global` connections so unsubscribe will clean them all up.

tgxworld:
Merging this in first as the changes are alot safer now.

PR #243, opened 2021-04-23
Allow tracing the performance of middleware

benlangfeld:
In a consuming application, when debugging slow responses to subscribing clients, it is useful to know where time is being spent in MessageBus middleware. This functionality allows hooking in a consumer-application-specific tracing mechanism, such as [NewRelic's `#trace_execution_scoped`](https://www.rubydoc.info/github/newrelic/newrelic-ruby-agent/NewRelic/Agent/MethodTracer#trace_execution_scoped-instance_method).

Tracking some attributes of a connection will allow us to spot patterns in use of long polling/chunked encoding, etc.

Our use case involves `NewRelic::Agent::MethodTracer.trace_execution_scoped` and `NewRelic::Agent.add_custom_attributes` but the implementation here is agnostic of the tool used to record this data.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,8 @@
 - Unrelease
 
+  - FEATURE: Permit tracing middleware performance via `MessageBus.tracer` and
+    `MessageBus#on_middleware_attributes`.
+
 02-10-2020
 
 - Version 3.3.4
--- a/README.md
+++ b/README.md
@@ -147,6 +147,33 @@ MessageBus.configure(on_middleware_error: proc do |env, e|
 end)
 ```
 
+### Performance monitoring
+
+MessageBus has hooks into its middleware for your application to perform monitoring of its performance.
+
+Firstly, the result of several decisions and the values used to arrive at them, such as whether to use long-polling or chunked encoding, can be provided to a logging routine of your definition:
+
+```ruby
+MessageBus.on_middleware_attributes do |attributes|
+  Rails.logger.debug(attributes)
+end
+```
+
+See `MessageBus.on_middleware_attributes` for the values that are provided.
+
+You can also specify a routine to wrap around various phases of the treatment of a request in order to measure its performance:
+
+```ruby
+MessageBus.tracer do |metric_name, &blk|
+  start_time = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+  blk.call
+  end_time = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+  Rails.logger.debug("Phase #{metric_name} took #{end_time - start_time} seconds")
+end
+```
+
+See `MessageBus.tracer` for the list of possible metric names. Don't forget to call the provided block.
+
 #### Disabling message_bus
 
 In certain cases, it is undesirable for message_bus to start up on application start, for example in a Rails application during the `db:create` rake task when using the Postgres backend (which will error trying to connect to the non-existent database to subscribe). You can invoke `MessageBus.off` before the middleware chain is loaded in order to prevent subscriptions and publications from happening; in a Rails app you might do this in an initializer based on some environment variable or some other conditional means.
--- a/lib/message_bus.rb
+++ b/lib/message_bus.rb
@@ -599,6 +599,46 @@ module MessageBus::Implementation
     @config[:client_message_filters]
   end
 
+  # Used to trace the performance of various phases of the treatment of a client request
+  # @yieldparam [String] metric_name some name for the metric being recorded. Possible values include:
+  #   * messagebus/middleware/authentication
+  #   * messagebus/middleware/subscriptions
+  #   * messagebus/middleware/headers
+  #   * messagebus/middleware/check_chunked
+  #   * messagebus/middleware/calculate_backlog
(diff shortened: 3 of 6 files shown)


rafbgarcia (approved):

benlangfeld:
Hey @SamSaffron , I hope you're well. If you are in agreement and were to have time to get this in a release, I would very much appreciate it :)

SamSaffron:
@jeremyevans I hope you have some time for a bit of advice here. 

I have a lot of general reticence of adding instrumentation that adds general cost to all execution (even when nobody is tracing) 

My general preference is to specify a "contract" for the various tracers out there, then the tracers can prepend specific methods or rewrite them safely. 

So if there is a registry saying: 

"message_bus#on_request" is in the class  MessageBus::Rack::Middleware #on_request we can define our own helper for rewriting these methods with instrumentation.


I am curious on your position on this problem of both what fidelity of instrumentation should be offered and implementation style. 

An interesting alternative may be to do a minor internal refactor and offer a "message-bus-tracing" gem.

jeremyevans:
My preference would be an approach that has zero or almost zero cost if not enabled, even if it significantly increases the cost when enabled.

This doesn't necessarily have to be a separate gem, we could do an internal refactor and keep it in the same gem, just have it be required separately.  That assumes we want to maintain the tracing ourselves. If we don't want to maintain it ourselves, we should have it be a separate gem.

Ideally, a user could just extend/include the necessary objects or classes to enable the tracing, having methods call `super` as appropriate, without having to modify the internals.  However, I'm not sure to what extent that is possible in this case.

benlangfeld:
@SamSaffron @jeremyevans Do you guys have any references to precedent that I could look at to come up with an implementation that satisfies your requirements?

jeremyevans:
To be clear, I don't have any requirements, only advice/recommendations.  In terms of the patch itself, my recommendation would be: instead of wrapping specific blocks of code with `@bus.trace`, extract private methods, so that those methods could be wrapped (calling `super`) to get the desired effect.  Example, replace:

```ruby
      user_id = @bus.user_id_lookup.call(env) if @bus.user_id_lookup
      group_ids = @bus.group_ids_lookup.call(env) if @bus.group_ids_lookup
      site_id = @bus.site_id_lookup.call(env) if @bus.site_id_lookup
```

with:

```
  user_id, group_ids, site_id = handle_lookups(env)
```

Then `handle_lookups` could be overridden by the tracing support.

Note that every extracted method has a small cost.  You would only want to extract a method to allow for additional flexibility if the benefit for the few users that need tracing is greater than the cost to all users that do not.

FWIW, in the above code example, if you wanted to trace, you could just have three separate traces, one inside each of the lookup procs.  That requires no code changes.

benlangfeld:
> To be clear, I don't have any requirements, only advice/recommendations. In terms of the patch itself, my recommendation would be: instead of wrapping specific blocks of code with `@bus.trace`, extract private methods, so that those methods could be wrapped (calling `super`) to get the desired effect. Example, replace:
> 
> ```ruby
>       user_id = @bus.user_id_lookup.call(env) if @bus.user_id_lookup
>       group_ids = @bus.group_ids_lookup.call(env) if @bus.group_ids_lookup
>       site_id = @bus.site_id_lookup.call(env) if @bus.site_id_lookup
> ```
> 
> with:
> 
> ```
>   user_id, group_ids, site_id = handle_lookups(env)
> ```
> 
> Then `handle_lookups` could be overridden by the tracing support.
> 
> Note that every extracted method has a small cost. You would only want to extract a method to allow for additional flexibility if the benefit for the few users that need tracing is greater than the cost to all users that do not.
> 
> FWIW, in the above code example, if you wanted to trace, you could just have three separate traces, one inside each of the lookup procs. That requires no code changes.

Thanks for the detail @jeremyevans ! By "could be overridden by the tracing support" do you mean simply having consuming applications subclass, or do you imagine something more involved?

jeremyevans:
Subclassing or including/prepending a module to the main class.

FWIW, my libraries that have need for extensibility (e.g. Sequel, Roda, Rodauth) have their own plugin systems to make things like this easier.  However, I'm not sure if that's a route that @SamSaffron wants to take, as that adds complexity too.  It's only worth it if there are sufficient plugins to justify it.  For example, if most of MessageBus's optional features were implemented as separate plugins, and users could pick only the plugins they need for their environment.

SamSaffron:
I love the plugin design @jeremyevans stuff has, but I worry it is a bit too big for what we need here. Maybe if we just prep for this longer term?

How does something like this look? 

https://gist.github.com/188f16bcd9a5efc7b2c9a74e14d833a5

I love the idea of having a "registry" of sorts cause it gives us lots of flexibility.

jeremyevans:
@SamSaffron Something like that should be fine.  The main change required to MessageBus would be extract methods for all behavior that you want to instrument.

SamSaffron:
Sorry Ben, closing this cause I would prefer a registry based approach here so we keep instrumentation further away from the code. 

https://gist.github.com/SamSaffron/188f16bcd9a5efc7b2c9a74e14d833a5#file-demo-rb-L27-L28

PR #244, opened 2021-04-26
Optimise responses to CORS preflight requests

benlangfeld:
Respond prior to authentication and subscription setup. Our response doesn't depend on these, so there's no point executing them, especially since they may be heavy routines which perform I/O.


SamSaffron (approved):

SamSaffron:
looks like a nice change to me!

PR #247, opened 2021-04-27
Remove cache busting implementation from AJAX shim

benlangfeld:
#246 intended to prevent this from happening, but we only tested with jQuery, whose default for this value is `true`. Our shim for `$.ajax` defaults to `false`, so in that context the desired change didn't take effect. This shim is supposed to be lightweight to address only MessageBus' requirements, so here we remove support for the now unused `cache` option entirely, rather than flipping the default.


SamSaffron:
Seems good to me! thanks Ben

PR #248, opened 2021-04-30
Allow logging the behaviour of middleware

benlangfeld:
In a consuming application, when debugging slow responses to subscribing clients, it is useful to know more detail about how message_bus handled the request.

Tracking some attributes of a connection will allow us to spot patterns in use of long polling/chunked encoding, compare response times to subscription volume, etc.

Our use case involves `NewRelic::Agent.add_custom_attributes` but the implementation here is agnostic of the tool used to record this data.

This was a part of https://github.com/discourse/message_bus/pull/243, and I think the less controversial part. I'm still working on getting something to satisfy the rest of that PR's feedback ready, but in the meantime maybe this can ship.

### DIFF
--- a/lib/message_bus/rack/middleware.rb
+++ b/lib/message_bus/rack/middleware.rb
@@ -139,6 +139,20 @@ class MessageBus::Rack::Middleware
 
     backlog = client.backlog
 
+    if @bus.on_middleware_attributes
+      @bus.on_middleware_attributes.call(
+        messagebus_seq: client.seq,
+        messagebus_query_string: env['QUERY_STRING'],
+        messagebus_client_count: @connection_manager.client_count,
+        messagebus_long_polling: long_polling,
+        messagebus_http_version: env['HTTP_VERSION'],
+        messagebus_dont_chunk: env['HTTP_DONT_CHUNK'],
+        messagebus_allow_chunked: allow_chunked,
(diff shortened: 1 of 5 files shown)


SamSaffron:
I follow, but I kind of want to wait on us establishing the pattern in  #243, it is possible a simple method extraction here would do the trick without needing to add the custom callback.

benlangfeld:
> I follow, but I kind of want to wait on us establishing the pattern in #243, it is possible a simple method extraction here would do the trick without needing to add the custom callback.

@SamSaffron Our tests revealed that our focus particularly should be within `Client#backlog`, and I don't personally anticipate needing to move forward with the rest of #243. I'll take a look at what changes are necessary to allow instrumentation of `Client#backlog`.

SamSaffron:
This has been hanging so long @benlangfeld, I am going to close this off for now.

lib/message_bus/rack/middleware.rb
```diff
@@ -139,6 +139,20 @@ def handle_request(env)
 
     backlog = client.backlog
 
+    if @bus.on_middleware_attributes
+      @bus.on_middleware_attributes.call(
+        messagebus_seq: client.seq,
```
SamSaffron, line 144: I think this is my biggest issue with this change, would prefer if we just pass env along and a fatter object vs all these options.

PR #250, opened 2021-05-06
FEATURE: support custom codecs for transport

SamSaffron:
Adds support for MessageBus.transport_codec which can be used to optimise
performance of transport

- Oj can be used as a serializer vs JSON which is a bit slower
- Users can implement custom codecs
- OjFast can be used as a serializer for efficient user_id lookups

Backwards compatible change, switching codecs to OjFast will invalidate all
existing storage.

Work in progress


SamSaffron:
Do not merge yet, work in progress

lib/message_bus/codec/oj_fast.rb
```diff
@@ -0,0 +1,81 @@
...
+      end
+
+      def initialize(packed)
+        @packed = packed
+      end
+
+      def include?(id)
+        found = (0...length).bsearch do |index|
+          start = index * 4
+          @packed[start, start + 4].unpack1("V") >= id
```
casperisfine, line 23: Hum, have you tried `@packed.byteslice(start..-1).unpack1("V")`, that should return you a shared string so avoid copy. 

But the current code only copy 4 bytes, so maybe it makes no difference.

SamSaffron:
good point trying it out... my initial implementation here is so buggy :)
getting there... slowly.

lib/message_bus/codec/oj_fast.rb
```diff
@@ -0,0 +1,82 @@
...
+      end
+
+      def include?(id)
+        found = (0...length).bsearch do |index|
+          @packed.byteslice(index * 4, 4).unpack1("V") >= id
+        end
+
+        if found
+          found && @packed.byteslice(found * 4, 4).unpack1("V") == id
+        end
```
rafbgarcia, line 28: TIL how to pack, thanks for educating me :)

Question though, you mentioned the problem with 4 bytes, but why is it not possible to use `"Q"` to pack 8 bytes?

I made a test and it seems to work:

```rb
numbers = [1,999999999999999].pack("Q*")
(0...numbers.length).step(8).map { |index| numbers.byteslice(index, 8).unpack1("Q") }
=> [1, 999999999999999]
```

Sorry if that doesn't make sense :)

jeremyevans:
Sorry for not being able to review the previous issue in a timely manner. I'm fine with custom codecs, as long as the default is still JSON for backwards compatibility.

SamSaffron:
@rafbgarcia 

See my latest changes, stuff should be reasonably tuned and well documented now. Let me know what you think.

@jeremyevans 

Absolutely, on top of that I am only shipping Codecs I consider 100% supported. (Oj and JSON) which are cross compatible.

Very interestingly @ohler55  JSON seems to outperform OJ for our typical use case cause we need to use Oj in compat mode which can be slower than JSON

lib/message_bus/codec/oj_fast.rb
SamSaffron: Thanks @casperisfine, I added this in! also fixed the terrible bug in that line

lib/message_bus/codec/oj_fast.rb
SamSaffron: Sure made Q vs V optional in the demo codec in the bench directory.

bench/codecs/packed_string.rb
```diff
@@ -0,0 +1,69 @@
...
+    end
+
+    def include?(id)
+      found = (0...length).bsearch do |index|
+        @packed.byteslice(index * @slot_size, @slot_size).unpack1(@pack_with) >= id
+      end
+
+      if found
+        found && @packed.byteslice(found * @slot_size, @slot_size).unpack1(@pack_with) == id
+      end
```
rafbgarcia, line 27: In case you care about it, this currently returns `true | nil`.

SamSaffron: yeah I noticed that ... will clean it up.

SamSaffron:
@rafbgarcia do you feel comfortable with where this is at? Should I go ahead and merge.

lib/message_bus/codec/oj.rb
```diff
@@ -0,0 +1,20 @@
...
+
+require 'oj' unless defined? ::Oj
+
+module MessageBus
+  module Codec
+    class Oj < Base
+      def initialize(options = { mode: :compat })
+        @options = options
+      end
+      def encode(hash)
```
rafbgarcia, line 11: minor, new line

```suggestion

      def encode(hash)
```

PR #258, opened 2021-12-04
FIX: prevent clobbering the session

### LINKED ISSUE #257: MessageBus simple polling clobbers the session cookie
I have traced a weird bug to what I believe is an unwanted behavior in MessageBus. What seems to happen is this:

1. The browser sends an AJAX request.
2. Server side, the controller handling the AJAX request sets new session values, and sends some MessageBus messages.
3. Before the AJAX request completes, MessageBus (browser side) starts a new polling request. This request contains the current (soon to become **OLD**) session cookie.
3. Now the AJAX request completes, setting the session cookie to a **NEW** value.
4. Finally, the MessageBus polling request completes, but the response contains the now **OLD** session cookie, hence clobbering the values set by the AJAX request.

Note that this does not seem to happen with long-polling, only when simple polling, as long polling did not include a "Set-Cookie" header. We can see on this screenshot that simple polling requests do contain a "Set-Cookies" header:

MessageBus_session_clobbering

I have not seen this reported anywhere, so maybe I am missing something or have things set up incorrectly. Anyway, MessageBus polling should not have side effects other than handling messages. In particular, it should not be sending session cookies, as this can clobber session state set by other means, like in the above scenario.

I have a tentative fix here: https://github.com/mediweb/message_bus/commit/f3db0f24d671cbd272e8ed836daad4267f7145cf

With this change in place, here is the result with the same scenario as above: simple polling responses do not have a "Set-Cookie" header anymore:

MessageBus_session_clobbering_fixed

Should I open a pull request, or am I missing something?

yemartin:
Simple polling was letting Rack commit the session, clobbering values
set by other means. Fix by telling Rack to skip committing the session.

Fixes #257

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,7 @@
+Unreleased
+
+  - FIX: Prevent simple polling from clobbering the session
+
 31-05-2021
 
 - Version 3.3.6
--- a/lib/message_bus/rack/middleware.rb
+++ b/lib/message_bus/rack/middleware.rb
@@ -66,6 +66,12 @@ class MessageBus::Rack::Middleware
   private
 
   def handle_request(env)
+    # Prevent simple polling from clobbering the session
+    # See: https://github.com/discourse/message_bus/issues/257
+    if (rack_session_options = env[Rack::RACK_SESSION_OPTIONS])
+      rack_session_options[:skip] = true
+    end
+
     # special debug/test route
     if @bus.allow_broadcast? && env['PATH_INFO'] == @broadcast_route
       parsed = Rack::Request.new(env)
--- a/spec/lib/message_bus/rack/middleware_spec.rb
+++ b/spec/lib/message_bus/rack/middleware_spec.rb
@@ -414,6 +414,11 @@ describe MessageBus::Rack::Middleware do
       JSON.parse(last_response.body).first["data"].must_equal('json' => true)
     end
 
+    it "should tell Rack to skip committing the session" do
+      post "/message-bus/1234", {}, {"rack.session.options" => {}}
+      last_request.env["rack.session.options"][:skip].must_equal true
+    end
+
     describe "on_middleware_error handling" do
       it "allows error handling of middleware failures" do
         @bus.on_middleware_error do |_env, err|


yemartin:
@SamSaffron @davidtaylorhq

It would be great if you could help write a better spec.

Note about the spec (copied here from #257 for better visibility): unfortunately I was unable to test the behavior (the presence of a "Set-Cookie" header), as I was not able to make the Rack session mechanism set a session cookie in specs. Extra setup is probably needed, but I am out of my depth here. So I fell back on testing the correct Rack session option being set in the request.

SamSaffron:
I think we are lucky here to have @jeremyevans as a contributor who knows Rack inside out. 

Jeremy is this the right thing to be doing?

jeremyevans:
I think this PR is probably better than the current behavior, assuming that message_bus is not modifying the session.  If there is a way for message_bus to modify the session in this middleware, however, this PR could affect backwards compatibility.

Note that this issue is not specific to message_bus.  I believe any use of rack's session support could be vulnerable to the same race condition for concurrent requests where the request that returns first modifies the session, as rack's session support does not check whether the current session data is already the same before setting the cookie (Roda's sessions support has this check, FWIW).

SamSaffron:
Tricky one @jeremyevans 

We support: 

```
MessageBus.user_id_lookup do |env|
    ... 
end
```

Technically you could do stuff with the environment and this change could break some session validation you did in MessageBus as a side effect. 


I guess the big question is if we consider this "breaking enough" to warrant a full dot release.  Or just consider this a minor change and doing "Session as Sideffect" to be a misuse.

I am leaning on letting this slip, your thoughts?

yemartin:
@SamSaffron Forgive me if it is obvious (it wasn't to me, I had to check the code): the `:skip` option only affects **committing** the session (i.e. writing the session cookie to the response headers). The rest of the session middleware behavior (extracting session ID, loading session data...) is unaffected and should keep working. So any session validation done in `MessageBus.user_id_lookup do |env| ...` should still work.

The only thing that could break is if someone was relying on **changing** the session from MessageBus. But the clobbering behavior seems to happen only when simple-polling (see #257). So while technically possible, since it would not work when long-polling, I doubt that anybody would be relying on this behavior.

Please @jeremyevans correct me if I am mistaken regarding the effect of the `:skip` option.

jeremyevans:
@yemartin is correct. I think this is fine to go in a minor or patch release.

SamSaffron (approved):

SamSaffron:
No worried, merged it in, looks like CI is in a bit of a mess, having a look

PR #260, opened 2021-12-08
Make message-bus-ajax use x-www-form-urlencoded Content-Type

jeremyevans:
I tried switching out jQuery for message-bus-ajax.js and it
didn't work.  After extensive debugging, I found that while
jQuery.ajax uses application/x-www-form-urlencoded,
message-bus-ajax.js was using JSON.  That broke my application,
which doesn't automatically parse JSON requests (it uses
roda-message_bus internally, which looks at the request body
before passing control to message_bus).  I think requests made by
message-bus-ajax.js should mirror the requests that jQuery.ajax
would use, so this switches message-bus-ajax.js to use
application/x-www-form-urlencoded.

This uses URLSearchParams to serialize the data.  Support for it
is pretty good (> 96%, about the same as CSS Grid).  If you want
to support a browser doesn't support URLSearchParams, you are
probably using jQuery anyway to handle cross-browser differences,
so you wouldn't be using message-bus-ajax.js.

CC @nathanstitt

### DIFF
--- a/assets/message-bus-ajax.js
+++ b/assets/message-bus-ajax.js
@@ -16,7 +16,7 @@
     for (var name in options.headers){
       xhr.setRequestHeader(name, options.headers[name]);
     }
-    xhr.setRequestHeader('Content-Type', 'application/json');
+    xhr.setRequestHeader('Content-Type', 'application/x-www-form-urlencoded');
     if (options.messageBus.chunked){
       options.messageBus.onProgressListener(xhr);
     }
@@ -31,7 +31,7 @@
         options.complete();
       }
     }
-    xhr.send(JSON.stringify(options.data));
+    xhr.send(new URLSearchParams(options.data).toString());
     return xhr;
   };
 


SamSaffron:
Seems fine to me, perhaps lets throw early if URLSearchParams is not available, will make it a bit easier to track down if it goes wrong cause it will be very early in the console messages.

I do wonder a bit if we should degrade a bit more gracefully and just log a warning and have it as a no-op so we don't destroy old clients, that said happy to leave as is.

jeremyevans:
I can add a warning up front if URLSearchParams isn't available.

In terms of degrading gracefully, do you mean use JSON if URLSearchParams is not available?  Not sure what was meant by "have it as a no-op".

SamSaffron:
Sorry Jeremy, was being a bit too brief here: 

1. Should we crash the JS loaders on old systems (which can happen if we do the check really early) ? 
2. Should we have message bus "crash" out with exceptions when it is subscribing, or silently eat the errors and log to console on old clients. 

Basically trying to figure out how friendly we should be to clients that don't support this. 

Could be all theoretical given how few unsupported clients there are.

SamSaffron:
I am find to take this as is, lets merge it as is. Thanks!

SamSaffron (approved):

PR #279, opened 2021-12-12
FIX: Stop pre-fork connection from being destroyed

CvX:
After forking, `Postgres::Client#subscribe` could fail in the parent with:

```
PQconsumeInput() server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
```

### DIFF
--- a/lib/message_bus/backends/postgres.rb
+++ b/lib/message_bus/backends/postgres.rb
@@ -44,6 +44,7 @@ module MessageBus
           @listening_on = {}
           @available = []
           @allocated = {}
+          @subscribe_connection = nil
           @mutex = Mutex.new
           @pid = Process.pid
         end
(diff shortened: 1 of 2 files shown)


SamSaffron:
Leaving this for @jeremyevans to review.

lib/message_bus/backends/postgres.rb
```diff
@@ -131,20 +132,23 @@ def subscribe(channel)
           listener = Listener.new
           yield listener
 
-          conn = raw_pg_connection
-          conn.exec "LISTEN #{channel}"
+          @subscribe_connection = raw_pg_connection
```
jeremyevans, line 135: Maybe just `conn = @subscribe_connection = raw_pg_connection`, to avoid unnecessary instance variable access and reduce the scope of the changes.

jeremyevans (approved):
This looks good to me.  See inline comments, but I have no objections to merging as-is.

PR #281, opened 2021-12-15
Remove unused variable

jeremyevans:
Found by running in verbose warnings mode.


SamSaffron:
thanks Jeremy, such an odd leftover. we are probably going to switch this transport to Rack:File, seems like the more consistent thing to do.

SamSaffron (approved):

PR #283, opened 2021-12-17
Remove files that no longer exist.

tgxworld:
This was breaking building of the gem with the following error:

```
WARNING:  See https://guides.rubygems.org/specification-reference/ for help
ERROR:  While executing gem ... (Gem::InvalidSpecificationException)
    ["vendor/assets/javascripts/message-bus-ajax.js", "vendor/assets/javascripts/message-bus.js"] are not files
```

### DIFF
--- a/message_bus.gemspec
+++ b/message_bus.gemspec
@@ -9,8 +9,7 @@ Gem::Specification.new do |gem|
   gem.summary       = %q{}
   gem.homepage      = "https://github.com/discourse/message_bus"
   gem.license       = "MIT"
-  gem.files         = `git ls-files`.split($\) +
-                      ["vendor/assets/javascripts/message-bus.js", "vendor/assets/javascripts/message-bus-ajax.js"]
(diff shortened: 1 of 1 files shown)


message_bus.gemspec
```diff
@@ -9,8 +9,7 @@ Gem::Specification.new do |gem|
   gem.summary       = %q{}
   gem.homepage      = "https://github.com/discourse/message_bus"
   gem.license       = "MIT"
-  gem.files         = `git ls-files`.split($\) +
-                      ["vendor/assets/javascripts/message-bus.js", "vendor/assets/javascripts/message-bus-ajax.js"]
+  gem.files         = `git ls-files`.split($\)
```
tgxworld, line 12: The `assets` folder is already returned by this command.

martin-brennan (approved):

CvX:
@tgxworld looks like it is used after all? https://github.com/discourse/discourse/runs/4576287009?check_suite_focus=true

PR #287, opened 2021-12-17
DEV: Update CI workflow

CvX:


gschlager (approved):

CvX:
Publishing the npm package will fail for now, since the NPM_TOKEN is not set yet. I'm merging it now anyway to publish the gem.

PR #289, opened 2021-12-18
DEV: Remove backend diagnostics

CvX:


CvX:
@benlangfeld How do you feel about dropping the diagnostics completely? That never really grew out into a full-fledged feature and seems to be unused by the vast majority of users. Its code was also untouched since #194. (three years ago 😃)

In the future, if there's any interest in better diagnostics interface, it could be created from the ground up as a separate gem/package.

SamSaffron (approved):

benlangfeld:
What was the motivation for this? Was it causing any problem?

benlangfeld:
Ah, now I see https://github.com/discourse/message_bus/security/advisories/GHSA-xmgj-5fh3-xjmm, I believe that is the motivation for this, correct?

CvX:
@benlangfeld Yes, that security issue was the initial motivation for dropping the diagnostics. It also wasn't being used enough (or at all) to justify keeping around this old code. And the ability to terminate server processes in production, even when correctly limited to just admins, was considered dangerous.

As I noted in the previous comment, it would be best to implement any future diagnostic tooling for message_bus as a separate, optional package.

PR #290, opened 2021-12-18
DEV: Various spec fixes

CvX:
1. DEV: Allow use of `TESTOPTS`
    Example:
    ```
    TESTOPTS="--verbose --seed=63387" bundle exec rake spec:postgres
    ```
2. DEV: Fix a ruby warning
3. DEV: Fix the error callback test
    10ms is not always enough for the task to finish. And queuing a task that triggers an error before you set the handler, doesn't really test anything (and in very rare cases it show the error message from the default handler)
4. DEV: Reset bus after multiproc spec to fix a leak


nattsw (approved):

nattsw:
Looks good but it's not obvious to me how this allows use of `TESTOPTS`

CvX:
It's the `Rake::TestTask` that takes care of that!

PR #291, opened 2021-12-19
DEV: Rename reliable_pub_sub to backend_instance

CvX:
It's already referred to as backend in most places, and the classes inherit from Backend::Base, so this makes the naming consistent.

Ideally I'd prefer `MessageBus#backend` instead of `MessageBus#backend_instance`, but `#backend` already exists so we can't rename both at the same time without breaking back-compatibility.


SamSaffron (approved):

SamSaffron:
Approved but it is probably worth doing a major version bump here once we make the diagnostics drop and this change.

PR #293, opened 2021-12-20
FIX: Restore dist js

### LINKED ISSUE #292: message-bus.js missing from vendor/assets/javascripts
`ActionView::Template::Error: couldn't find file 'message-bus' with type 'application/javascript'` 

We've just upgraded this gem in our Rails app (6.1.4.1) from 3.3.1 to 3.3.7 and we're getting this error. I'm going to look through your commits to see if I can figure out why or if it's something our end but just opening this issue in case anyone's spotted it in the meantime.

CvX:
Fixes #292


svpersteve:
This was super quick, thanks! 🙏🏻

PR #306, opened 2022-01-16
FIX: Make pg table creation thread-safe

CvX:
This hopefully fixes spec flakes:

```
ERROR:  duplicate key value violates unique constraint "pg_class_relname_nsp_index" (PG::UniqueViolation)
DETAIL:  Key (relname, relnamespace)=(message_bus_id_seq, 2200) already exists.
```

```
ERROR:  relation "message_bus" does not exist (PG::UndefinedTable)
LINE 1: SELECT 'message_bus'::regclass
```

(See: https://github.com/discourse/message_bus/runs/4772136781?check_suite_focus=true)

### DIFF
--- a/lib/message_bus/backends/postgres.rb
+++ b/lib/message_bus/backends/postgres.rb
@@ -175,9 +175,16 @@ module MessageBus
         end
 
         def create_table(conn)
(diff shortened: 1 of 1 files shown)


SamSaffron:
@jeremyevans are you good with this... technically this add thread safety which is hand for tests but we do not have multi process safety which we may want to consider here and possibly in other places?

jeremyevans:
This seems to improve what we have, so I'm OK with it.  Maybe we could add multiprocess safety by rescuing the CREATE TABLE failure and not doing the index creation in that case, instead checking again that the table exists?

SamSaffron (approved):

PR #308, opened 2022-02-16
PERF: Optimize `Client#backlog` for up-to-date clients

davidtaylorhq:
`Client#backlog` is most commonly used by the HTTP server. In long-polling mode clients send the id of the last message they received on each channel. They are then given the 'backlog' of more recent messages, and are subscribed to any further messages for the rest of the HTTP poll. In the vast majority of cases, there are no new messages available immediately, so the backlog is empty.

Previously, `Client#backlog` would make a separate `MessageBus#backlog` call for every subscribed channel. Under some circumstances it would make further calls to `MessageBus#last_id` for each channel. These calls scale linearly with the number of channels a user is subscribed to.

This commit refactors `Client#backlog` so that it starts by making a request to the backend for the 'last_ids' of all requested channels. Using this information, it can decide whether there is any need to call the more expensive `MessageBus#backlog` method. In most cases, there is no need.

A new `last_ids(*channels)` method is introduced which allows fetching multiple `last_id` values simultaneously. Implementations for all three backends have been added. Redis uses `MGET`, Postgres uses a single SELECT call, and the Memory backend leans on the existing #last_id method.

`spec/performance/backlog.rb` was written to test this change. On a machine with local redis/postgres, it is up to 8x faster for the (arguably most common) case where the client is already up-to-date. It becomes slightly slower once > 8 channels have backlogged messages.

Given the reduced number of round-trips, improvements may be even greater in a production environment with a remote database.

---

Time taken to process 10,000 backlog calls for a client subscribed to 10 channels. Before and after this commit:

Redis:
```
                                                            (before)   (after)
redis - #backlog with no backlogs requested                  2.53 s    0.63 s
redis - #backlog when 0/10 channels have new messages        5.05 s    0.62 s
redis - #backlog when 2/10 channels have new messages        4.79 s    1.31 s
redis - #backlog when 4/10 channels have new messages        4.47 s    2.06 s
redis - #backlog when 6/10 channels have new messages        4.16 s    2.68 s
redis - #backlog when 8/10 channels have new messages        3.84 s    3.35 s
redis - #backlog when 10/10 channels have new messages       3.47 s    4.04 s
```

Postgres:
```
                                                           (before)   (after)
postgres - #backlog with no backlogs requested               2.21 s    1.75 s
postgres - #backlog when 0/10 channels have new messages     5.38 s    1.73 s
postgres - #backlog when 2/10 channels have new messages     5.05 s    2.24 s
postgres - #backlog when 4/10 channels have new messages     4.71 s    2.70 s
postgres - #backlog when 6/10 channels have new messages     4.41 s    3.14 s
postgres - #backlog when 8/10 channels have new messages     4.04 s    3.59 s
postgres - #backlog when 10/10 channels have new messages    3.68 s    4.05 s
```

### DIFF
--- a/lib/message_bus.rb
+++ b/lib/message_bus.rb
@@ -512,7 +512,12 @@ module MessageBus::Implementation
   def last_ids(*channels, site_id: nil)
     encoded_channel_names = channels.map { |c| encode_channel_name(c, site_id) }
     ids = backend_instance.last_ids(*encoded_channel_names)
-    channels.zip(ids).to_h
+
+    result = {}
+    channels.each_with_index do |channel, index|
+      result[channel] = ids[index]
+    end
+    result
   end
 
   # Get the last message published on a channel
--- a/lib/message_bus.rb
+++ b/lib/message_bus.rb
@@ -503,6 +503,18 @@ module MessageBus::Implementation
     backend_instance.last_id(encode_channel_name(channel, site_id))
   end
 
+  # Get the ID of the last message published on multiple channels
+  #
+  # @param [Array<String>] channels - array of channels to fetch
+  # @param [String] site_id - the ID of the site by which to filter
+  #
+  # @return [Array<Integer>] the channel-specific IDs of the last message published to each requested channel
+  def last_ids(*channels, site_id: nil)
+    encoded_channel_names = channels.map { |c| encode_channel_name(c, site_id) }
+    ids = backend_instance.last_ids(*encoded_channel_names)
+    channels.zip(ids).to_h
+  end
+
   # Get the last message published on a channel
   #
   # @param [String] channel the name of the channel in question
--- a/lib/message_bus.rb
+++ b/lib/message_bus.rb
@@ -508,7 +508,7 @@ module MessageBus::Implementation
   # @param [Array<String>] channels - array of channels to fetch
   # @param [String] site_id - the ID of the site by which to filter
   #
-  # @return [Array<Integer>] the channel-specific IDs of the last message published to each requested channel
+  # @return [Hash] the channel-specific IDs of the last message published to each requested channel
   def last_ids(*channels, site_id: nil)
     encoded_channel_names = channels.map { |c| encode_channel_name(c, site_id) }
     ids = backend_instance.last_ids(*encoded_channel_names)
--- a/lib/message_bus.rb
+++ b/lib/message_bus.rb
@@ -512,12 +512,7 @@ module MessageBus::Implementation
   def last_ids(*channels, site_id: nil)
     encoded_channel_names = channels.map { |c| encode_channel_name(c, site_id) }
     ids = backend_instance.last_ids(*encoded_channel_names)
-
-    result = {}
-    channels.each_with_index do |channel, index|
-      result[channel] = ids[index]
-    end
-    result
+    channels.zip(ids).to_h
   end
 
   # Get the last message published on a channel
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,12 @@
+16-02-2022
+
+- Version 4.1.0
+
+  - PERF: Optimize Client#backlog for up-to-date clients
+
+    Also introduces a new MessageBus#last_ids(*channels) api for fetching the last_ids of
+    multiple channels simultaneously
+
 11-01-2022
 
 - Version 4.0.0
--- a/lib/message_bus/backends/base.rb
+++ b/lib/message_bus/backends/base.rb
@@ -118,6 +118,15 @@ module MessageBus
         raise ConcreteClassMustImplementError
       end
 
+      # Get the ID of the last message published on multiple channels
+      #
+      # @param [Array<String>] channels - array of channels to fetch
+      #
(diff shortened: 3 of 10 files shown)


lib/message_bus.rb
```diff
@@ -503,6 +503,18 @@ def last_id(channel, site_id = nil)
...
+  # Get the ID of the last message published on multiple channels
+  #
+  # @param [Array<String>] channels - array of channels to fetch
+  # @param [String] site_id - the ID of the site by which to filter
+  #
+  # @return [Array<Integer>] the channel-specific IDs of the last message published to each requested channel
+  def last_ids(*channels, site_id: nil)
+    encoded_channel_names = channels.map { |c| encode_channel_name(c, site_id) }
+    ids = backend_instance.last_ids(*encoded_channel_names)
+    channels.zip(ids).to_h
```
SamSaffron, line 515: this can be micro optimized if you avoid allocating the temporary array

SamSaffron:
Very supportive of this change, can you expand on why the 10/10 case gets slower?

lib/message_bus.rb
davidtaylorhq: As in, something like this?

```
def last_ids(*channels, site_id: nil)
    ids = backend_instance.last_ids(*channels.map { |c| encode_channel_name(c, site_id) })
    channels.zip(ids).to_h
end

**Comment by davidtaylorhq** [comment]
> can you expand on why the 10/10 case gets slower?

Taking Redis as an example.

Previously the 10/10 case would result in `10x ZRANGEBYSCORE` calls. Now, it is `1x MGET + 10x ZRANGEBYSCORE`.

That extra MGET is a constant cost, so it shouldn't get much worse as the number of channels increases.

But in the optimum (and I think, probably the most common) 0/10 case, we've replaced `10x ZRANGEBYSCORE + 10x GET` calls with a single `MGET` call

File: lib/message_bus.rb [hunk 2]
```diff
@@ -503,6 +503,18 @@ def last_id(channel, site_id = nil)
     backend_instance.last_id(encode_channel_name(channel, site_id))
   end
 
+  # Get the ID of the last message published on multiple channels
+  #
+  # @param [Array<String>] channels - array of channels to fetch
+  # @param [String] site_id - the ID of the site by which to filter
+  #
+  # @return [Array<Integer>] the channel-specific IDs of the last message published to each requested channel
```
ZogStriP, line 511: Isn't it returning a `Hash`?

```suggestion
  # @return [Hash] the channel-specific IDs of the last message published to each requested channel
```

lib/message_bus.rb
CvX: ^ that would still be the same. I think Sam meant the array that's the result of `#zip`, which can be avoided with something like:

```rb
channels.each.with_index.with_object({}) do |channel, index, result|
  result[channel] = ids[index]
end
```

…but I don't know how much more performant that is 🙂

eviltrout (approved):

lib/message_bus.rb
```diff
@@ -503,6 +503,23 @@ def last_id(channel, site_id = nil)
...
+  #
+  # @param [Array<String>] channels - array of channels to fetch
+  # @param [String] site_id - the ID of the site by which to filter
+  #
+  # @return [Hash] the channel-specific IDs of the last message published to each requested channel
+  def last_ids(*channels, site_id: nil)
+    encoded_channel_names = channels.map { |c| encode_channel_name(c, site_id) }
+    ids = backend_instance.last_ids(*encoded_channel_names)
+
+    result = {}
```
ZogStriP, line 516: I'm curious, did you benchmark this change? Is it faster/lower on memory?

ZogStriP (approved):

lib/message_bus.rb
davidtaylorhq: I didn't... but I guess I should

```
require 'benchmark/ips'

Benchmark.ips do |x|
  channels = ["one", "two", "three", "four", "five"]
  ids = [1, 2, 3, 4, 5]

  x.report("zip") do 
    channels.zip(ids).to_h
  end

  x.report("each_with_index") do
    result = {}
    channels.each_with_index do |channel, index|
      result[channel] = ids[index]
    end
  end

  x.report("while") do
    result = {}
    i = 0
    while i < channels.size
      result[channels[i]] = ids[i]
      i += 1 
    end
  end

  x.report("transpose") do
    [channels, ids].transpose.to_h
  end

  x.compare!
end
```

```
Comparison:
               while:  1780741.1 i/s
           transpose:  1761829.2 i/s - same-ish: difference falls within error
                 zip:  1737068.4 i/s - same-ish: difference falls within error
     each_with_index:  1531362.7 i/s - 1.16x  (± 0.00) slower
```

😅 Looks like zip was faster after all. Will revert that commit

(ultimately this is negligible compared to the network round-trip for redis, so I'm not too worried)

discoursereviewbot: [SamSaffron posted](https://review.discourse.org/t/perf-optimize-client-backlog-for-up-to-date-clients-pr-308/31812/10):

I see yeah erasing the array alloc is kind of pointless then

PR #314, opened 2022-10-24
DEV: Assume there's `progress` event support

CvX:
I'd say that support for `XMLHttpRequest`'s `progres` event is rather wide (Chrome 1, Firefox 1, iOS Safari 1, etc.) See: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/progress_event#browser_compatibility

(also: fix typos)

### DIFF
--- a/assets/message-bus.js
+++ b/assets/message-bus.js
@@ -88,11 +88,6 @@
     return false;
   };
 
-  var hasonprogress = new XMLHttpRequest().onprogress === null;
-  var allowChunked = function () {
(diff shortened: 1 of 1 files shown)


SamSaffron:
Hmmm so ... drop IE9 support for message_bus

https://caniuse.com/mdn-api_xmlhttprequest_progress_event

Commit message needs to be more accurate, I guess I am Ok with it. I think we auto disable chunking anyway if we detect it is not working due to proxy etc...

PR #315, opened 2022-10-24
DEV: Assume `console` is present

CvX:
And use `console.error` where needed.


SamSaffron:
do not agree with this, fine though if you check for console.error 

message_bus supports quite old browsers.

PR #317, opened 2022-10-25
Allow publishing after turning the bus off

benlangfeld:
Many use cases for disabling the bus only really care about the subscription side, and still need to be able to publish.

See also #185.

### DIFF
--- a/spec/lib/message_bus_spec.rb
+++ b/spec/lib/message_bus_spec.rb
@@ -17,26 +17,6 @@ describe MessageBus do
     @bus.destroy
   end
 
-  it "can be turned off" do
-    @bus.off
-
-    @bus.off?.must_equal true
-  end
-
-  it "can call destroy multiple times" do
-    @bus.destroy
-    @bus.destroy
-    @bus.destroy
-  end
-
-  it "can be turned on after destroy" do
-    @bus.destroy
-
-    @bus.on
-
-    @bus.after_fork
-  end
-
   it "destroying immediately after `after_fork` does not lock" do
     10.times do
       @bus.on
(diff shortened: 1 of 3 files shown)


SamSaffron:
I support this change but we need a test to back it, can you add it in?

spec/lib/message_bus_spec.rb
```diff
@@ -216,6 +196,40 @@
...
+  end
+
+  it "can be turned off only for subscriptions" do
+    @bus.off(disable_publish: false)
+
+    @bus.off?.must_equal true
+
+    @bus.publish("/chuck", "norris")
+
+    @bus.backlog("/chuck").map(&:data).to_a.must_equal ["norris"]
```
SamSaffron, line 216: this is close but a bit confusing, you should be attempting a new subscription and that should fail, seems to only be testing part of the change?

benlangfeld: If I do an explicit subscription after calling `#off`, the subscription works. `#off` only affects the implicit middleware subscription.

benlangfeld: What do you think of https://github.com/discourse/message_bus/pull/317/commits/45fd7387230e705938fa74e901af235c15fc1f64 ?

spec/lib/message_bus_spec.rb
```diff
@@ -216,6 +196,50 @@
...
+
+    @bus.subscribe("/chuck") do |msg|
+      data = msg.data
+    end
+
+    @bus.publish("/chuck", "norris")
+
+    @bus.backlog("/chuck").map(&:data).to_a.must_equal ["norris"]
+
+    sleep 2
```
SamSaffron, line 224: this is tricky, need to figure out how to do this without the sleep... maybe we tag as a memory only test and then we can control the time better. 

This is not an easy feature to test, for sure.

SamSaffron:
sorry been snowed under today, going to have a think about it and have a
look tomorrow, I really don't want to add the sleep into the test suite,
that is for sure

SamSaffron (approved):

SamSaffron:
Yeah this looks better, thanks! Merging it in.

PR #318, opened 2022-10-25
Redis 5.0 gem is incompatible with our init params

benlangfeld:
See https://github.com/redis/redis-rb/issues/1159

### DIFF
--- a/message_bus.gemspec
+++ b/message_bus.gemspec
@@ -19,7 +19,7 @@ Gem::Specification.new do |gem|
   gem.add_runtime_dependency 'rack', '>= 1.1.3'
 
   # Optional runtime dependencies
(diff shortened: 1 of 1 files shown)


SamSaffron:
I would prefer to fix in a cross compat way, this sweeps under the carpet a bit ...

benlangfeld:
> I would prefer to fix in a cross compat way, this sweeps under the carpet a bit ...

I also would prefer to do that, but I'm on a deadline with https://github.com/discourse/message_bus/pull/317 and am prioritising being able to keep moving.

SamSaffron (approved):

SamSaffron:
approving for now, will see if I can fix Redis 5 as well.

PR #319, opened 2022-10-26
FIX: Add Redis 5 support

SamSaffron:
Redis 5 is picky about the params you send it, in 4 and below you could ship
it arbitrary params and it would simply ignore. In 5 it requires specific ones.

The change also removes a non relevant test case, Redis::Client::Connector
is no longer a thing Redis gem version 5 and up.

Redis 5 gem also duplicates incoming params so it is inherently more safe.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,6 @@
 FUTURE
 
-  - FIX: Add redis version 5 support
+  - FIX: Add redis gem version 5 support
 
 22-02-2022
 
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,7 @@
+FUTURE
+
+  - FIX: Add redis version 5 support
+
 22-02-2022
 
 - Version 4.2.0
--- a/lib/message_bus/backends/redis.rb
+++ b/lib/message_bus/backends/redis.rb
@@ -346,7 +346,29 @@ LUA
       private
 
       def new_redis_connection
-        ::Redis.new(@redis_config.dup)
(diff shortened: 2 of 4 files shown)


lib/message_bus/backends/redis.rb
```diff
@@ -346,7 +346,29 @@ def global_subscribe(last_id = nil, &blk)
       private
 
       def new_redis_connection
-        ::Redis.new(@redis_config.dup)
+        config = @redis_config.filter do |k, v|
+          # This is not ideal, required for Redis gem version 5
+          # redis-client no longer accepts arbitrary params
+          # anything unknown will error out.
+          # https://github.com/redis-rb/redis-client/blob/master/lib/redis_client/config.rb#L21-L39
```
tgxworld, line 353: Minor nit but we almost never want to point to a branch when leaving links to github because the branch can update in the future and the line numbers may not be accurate. You can get the permanent link if you press `y` while viewing the file on github.

CHANGELOG
```diff
@@ -1,3 +1,7 @@
+FUTURE
+
+  - FIX: Add redis version 5 support
```
tgxworld, line 3:
```suggestion
  - FIX: Add redis rubygem version 5 support
```

We need to be clearer here and with the commit message because `Redis 5` may sound like we're referring to Redis instead of the `redis` rubygem.

tgxworld (approved):
I left some notes about the commit message and comments but implementation looks OK.

PR #324, opened 2023-01-05
FIX: Ensure non-long-polling requests are always spaced out

davidtaylorhq:
When long-polling, after a request returns some data, message-bus will immediately begin another long-polling request to maintain a consistent stream of messages.

When not long polling, we do not expect a continuous stream of messages, and the client aims to make one request every `backgroundCallbackInterval` (default 60s).

Before this commit, some long-polling logic was leaking into the non-long-polling requests. Under certain situations (e.g. a channel receiving very frequent messages), this could cause clients to start issuing non-long-polling requests every `minPollInterval` (100ms). This commit fixes the logic so that 'immediate retry' is only performed when in long-polling mode.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -11,4 +11,13 @@ module.exports = {
   },
   rules: {},
   ignorePatterns: ['/vendor', '/doc', '/assets/jquery-1.8.2.js'],
+  overrides: [
+    {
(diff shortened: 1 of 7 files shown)


CvX (approved):

ZogStriP (approved):
LGTM

PR #325, opened 2023-01-06
FIX: Do not disable chunking on cancel/error

davidtaylorhq:
If the first chunk of a message takes more than 3s, we assume an intermediate proxy is buffering data and drop into dont-chunk mode. Previously, this 3s timer would continue even if a request was aborted before the first chunk arrived (e.g. channel subscriptions change in quick succession). This commit ensures the timer is cancelled when the request is aborted.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,12 @@
 FUTURE
 
+
+13-01-2023
+
+- Version 4.2.1
+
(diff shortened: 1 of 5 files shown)


davidtaylorhq:
(need to make the tests more robust for this)

nbianca (approved):

PR #330, opened 2023-04-28
Fix bad exclusion of messagebus configuration option in redis connection

Intrepidd:
Creating just a DRAFT PR for now to get insights.

I discovered it is impossible to use the group ids lookup feature alongside the redis backend as there is a typo in the exclusion list.

As stated by the comment, we should probably allowlist the params to be passed to redis rather then exclude some known ones. Maybe even better would be to have them in a separate `redis_options` hash but this would kill backwards compatibility.

Additionally, a regression spec is needed

Any idea which direction is preferred ?

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,7 @@
+FUTURE
+
+- FIX: Typo in `group_ids_lookup` option name in Redis params exclusion list that was preventing the use of the redis backend along with this option.
+
 28-06-2023
 
 - Version 4.3.7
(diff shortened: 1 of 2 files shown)


Intrepidd:
@tgxworld Sorry to ping you on this one. I noticed you merged a similar PR 

Are you fine with me just adding a changelog entry to be able to merge this PR ?

tgxworld:
Yes please :) Do add a changelog here and I'll be happy to merge this.

Intrepidd:
Done !

tgxworld (approved):

tgxworld:
Thank you @Intrepidd

Intrepidd:
Thanks! Do you think this is worth cutting a patch level release?

PR #331, opened 2023-04-28
Update blurb on middleware stack

Intrepidd:
There's a much easier way since Rails 6.1.3


Drenmi (approved):
Thanks, @Intrepidd! 🙇

PR #336, opened 2023-06-19
FIX: Don't error when trying to write a message to a closed client

tgxworld:
When a client is set to use chunked encoding by setting `use_chunked` to
true, trying to write a message to the client when the client has been
closed will raise an error. This commit changes it such that we don't
raise an error since the raised error doesn't do anything useful.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,10 +1,13 @@
-FUTURE
+19-06-2023
+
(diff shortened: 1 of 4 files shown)


SamSaffron (approved):

tgxworld:
Thank you for reviewing @SamSaffron

PR #347, opened 2024-01-16
DOC: Fix CHANGELOG version numbering for 4.3.0 and 4.3.1

sobstel:
It looks like 4.3.0 and 4.3.1 were wrongly labeled as 4.2.0 and 4.2.1 respectively in CHANGELOG file.

It can be confusing, eg I thought that it's enough to update to 4.2.0 to have redis gem ver 5 support.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -42,13 +42,13 @@
 
 06-01-2023
 
-- Version 4.2.1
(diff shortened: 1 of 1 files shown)


CvX (approved):

CvX:
Thanks! 🙂

PR #353, opened 2024-06-19
DEV: Fix linting

CvX:


CvX:
Yup, that's expected. But after consideration I decided to extract linting so it runs once, on a supported ruby version.

PR #388, opened 2026-04-10
fix: subscriber thread never self heals after half open tcp connection

### LINKED ISSUE #387: Global messages on XX timed out, subscriber thread never self-heals after half-open TCP connection
When a Unicorn/Puma worker logs:

> Global messages on timed out, message bus is no longer functioning correctly

the process never recovers on its own. Real-time features silently break for all users on that worker until the app server is restarted manually.

MessageBus library only logs a warning but there is no revive mechanism.

I have been using MessageBus with Puma and deployed 2 different applications on 2 different AWS accounts in 2 different regions and when it happens in 1 application it happens to the other as well. That's why I suspect this is something happening at the network layer (like a half-open TCP connection).

I know that in Discourse you use Unicorn and you have a programmatic way to revive things when something goes wrong. Or maybe you don't deploy on AWS and have never experienced this network issue that puts the library in a "dead" state. But on my side, with Puma and deploying on AWS EC2 servers, I had to manually restart the web server each time it happens. (Most of the time this is during the night and by the time I wake up I already have many tickets on my app complaining about live updates 😄 ). It is also pretty random, it can "not happen for weeks", then happen 2 days in a row.

I made a fix that I tested for a long time (using my fork version of MessageBus in the Gemfile) to make sure there were no side effects. I waited for the warning to happen again and for the fix to "revive" the message bus without any side issues. I monitored it for a couple of days where it happened 2 times, and both times it recovered cleanly.

Pull request coming.

Related to these 2 Discourse threads:

- https://meta.discourse.org/t/message-bus-is-no-longer-functioning-correctly/237209
- https://meta.discourse.org/t/after-config-change-some-settings-not-in-sync-on-sibling-nodes-in-multi-node-discourse-setup/372573/4?tl=en

Mubramaj:
Closes #387 

## Problem

When a Unicorn/Puma worker logs:

> Global messages on timed out, message bus is no longer functioning correctly

the process never recovers. Real-time features silently break for all users on that worker until the app server is restarted manually.

This happens in practice when Redis and MessageBus run on separate hosts and a half-open TCP connection forms (common after VRRP failover or transient network events). The subscriber's `global_redis.subscribe` call blocks on a dead socket indefinitely without raising an exception. The thread stays alive (no exception, so `thread.alive? == true`) but receives no messages, so `@last_message` goes stale. `ensure_subscriber_thread` sees a live thread and does nothing.

MessageBus detects the stale `@last_message` via the keepalive mechanism and logs a warning but stops there.

### The Fix

When the keepalive detects a stale `@last_message`, kill the stuck subscriber thread and start a fresh one with a new Redis connection:

```ruby
# new behaviour
thread.kill
@subscriber_thread = nil
ensure_subscriber_thread  # opens a new connection
```
I found out that @SamSaffron made a commit a long time ago that removed an original self-kill mechanism https://github.com/discourse/message_bus/commit/60eda3d 

`60eda3d` removed the original self-kill mechanism because it could take down a healthy worker during the network issue. Here we just kill the thread and revive using `ensure_subscriber_thread` . Sam let me know if I am totally missing something and re-introducing an issue. 

I ran this fix in production via a fork-pinned Gemfile for several days. The warning occurred twice; both times the subscriber thread restarted cleanly with no side effects.

1. Happened last Thursday a 7:04 am 
image

2. Then again 2 times this morning 

image

In both cases it revived cleanly. 

### Related

- https://meta.discourse.org/t/message-bus-is-no-longer-functioning-correctly/237209
- https://meta.discourse.org/t/after-config-change-some-settings-not-in-sync-on-sibling-nodes-in-multi-node-discourse-setup/372573/4?tl=en

### DIFF
--- a/lib/message_bus.rb
+++ b/lib/message_bus.rb
@@ -752,9 +752,20 @@ module MessageBus::Implementation
         publish("/__mb_keepalive__/", Process.pid, user_ids: [-1])
         if (Time.now - (@last_message || Time.now)) > keepalive_interval * 3
           logger.warn "Global messages on #{Process.pid} timed out, message bus is no longer functioning correctly"
+          # The subscriber thread is stuck (half-open TCP connection or similar network issue)
+          # and is no longer receiving messages. Kill it and let ensure_subscriber_thread
+          # create a fresh one with a new Redis connection.
+          @mutex.synchronize do
+            if @subscriber_thread == thread
+              thread.kill
+              @subscriber_thread = nil
+            end
+          end
+          ensure_subscriber_thread
+          # The new thread sets up its own keepalive blk stop re-queuing this one.
+        else
+          timer.queue(keepalive_interval, &blk) if keepalive_interval > MIN_KEEPALIVE
         end
-
-        timer.queue(keepalive_interval, &blk) if keepalive_interval > MIN_KEEPALIVE
       end
     end
 
--- a/spec/lib/message_bus_spec.rb
+++ b/spec/lib/message_bus_spec.rb
@@ -398,4 +398,85 @@ describe MessageBus do
       channel.must_equal('/test')
     end
   end
+
+  describe "keepalive recovery" do
+    # MIN_KEEPALIVE is 20s in production, making real-timer tests take 60s+.
+    # We temporarily lower it so keepalive_interval=1 is accepted, which
+    # triggers recovery at ~4s fast enough for CI, slow enough to be real.
+    FAST_KEEPALIVE = 1
+
+    before do
+      @original_min_keepalive = MessageBus::Implementation::MIN_KEEPALIVE
+      MessageBus::Implementation.send(:remove_const, :MIN_KEEPALIVE)
+      MessageBus::Implementation.const_set(:MIN_KEEPALIVE, 0)
+
+      @log_output = StringIO.new
+      @bus.configure(keepalive_interval: FAST_KEEPALIVE, logger: Logger.new(@log_output))
+
+      # Capture the real backend method before stubbing.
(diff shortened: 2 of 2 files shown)


SamSaffron:
The trouble is that `Thread.kill` is fundamentally unsafe and can lead to many other issues. My concern is that the cure being proposed is more dangerous than the illness. 

The only safe thing to do is tear down the process and start again or rely on other timeouts that are safe to clean stuff up. 

@jeremyevans thoughts?

Mubramaj:
Thank you for the feedback and the insight. I totally understand your concern and the issue. Based on your feedback I tried a different route:


Opened a PR on my fork and going to test it again in production for couple days until the illness appears again and there is no side effect. Then I will open a PR again. But in case you have right away feedback on the new route taken here is the link:

https://github.com/Mubramaj/message_bus/pull/2/changes 

Basically the idea is instead of killing the thread simulate an error by disconnecting the back-end and for the existing global_subscribe   `rescue => error ... retry` mechanism kicks in. 

This way the subscriber thread never dies. It reconnects itself. No Thread.kill.

jeremyevans:
> The trouble is that `Thread.kill` is fundamentally unsafe and can lead to many other issues. My concern is that the cure being proposed is more dangerous than the illness.
> 
> The only safe thing to do is tear down the process and start again or rely on other timeouts that are safe to clean stuff up.
> 
> @jeremyevans thoughts?

I agreed that we should not use `Thread.kill`. https://github.com/Mubramaj/message_bus/pull/2/changes looks like a better way to fix this.

Mubramaj:
Thank you @jeremyevans for the insight. 

I will test it on production and once I validate it fixes the issue will clean up my branch and open a new PR. 

Should I close this one ?

PR #389, opened 2026-04-22
fix: subscriber thread never self heals after connection issue

### LINKED ISSUE #387: Global messages on XX timed out, subscriber thread never self-heals after half-open TCP connection

Mubramaj:
follow up on https://github.com/discourse/message_bus/pull/388 

Before opening this PR I first tested it on my production setup that was experiencing the "Global messages" timeout issues to make sure it fixes it without any side effect. 

Happened this morning and self healed correctly:

First warning message at 7:17 am

`Global messages on 1 timed out, message bus is no longer functioning correctly`

Followed by another warning

`stream closed in another thread subscribe failed, reconnecting in 1 second.` (7:17 am as well ) 

It self healed correctly didn't have to restart the rails app . 

Closes #387

### DIFF
--- a/lib/message_bus.rb
+++ b/lib/message_bus.rb
@@ -752,9 +752,20 @@ module MessageBus::Implementation
         publish("/__mb_keepalive__/", Process.pid, user_ids: [-1])
         if (Time.now - (@last_message || Time.now)) > keepalive_interval * 3
           logger.warn "Global messages on #{Process.pid} timed out, message bus is no longer functioning correctly"
+          # The subscriber thread is stuck (half-open TCP connection or similar network issue)
+          # and is no longer receiving messages. Kill it and let ensure_subscriber_thread
+          # create a fresh one with a new Redis connection.
+          @mutex.synchronize do
+            if @subscriber_thread == thread
+              thread.kill
+              @subscriber_thread = nil
+            end
+          end
+          ensure_subscriber_thread
+          # The new thread sets up its own keepalive blk stop re-queuing this one.
+        else
+          timer.queue(keepalive_interval, &blk) if keepalive_interval > MIN_KEEPALIVE
         end
-
-        timer.queue(keepalive_interval, &blk) if keepalive_interval > MIN_KEEPALIVE
       end
     end
 
--- a/lib/message_bus.rb
+++ b/lib/message_bus.rb
@@ -752,20 +752,14 @@ module MessageBus::Implementation
         publish("/__mb_keepalive__/", Process.pid, user_ids: [-1])
         if (Time.now - (@last_message || Time.now)) > keepalive_interval * 3
           logger.warn "Global messages on #{Process.pid} timed out, message bus is no longer functioning correctly"
-          # The subscriber thread is stuck (half-open TCP connection or similar network issue)
-          # and is no longer receiving messages. Kill it and let ensure_subscriber_thread
-          # create a fresh one with a new Redis connection.
-          @mutex.synchronize do
-            if @subscriber_thread == thread
-              thread.kill
-              @subscriber_thread = nil
-            end
-          end
-          ensure_subscriber_thread
-          # The new thread sets up its own keepalive blk stop re-queuing this one.
-        else
-          timer.queue(keepalive_interval, &blk) if keepalive_interval > MIN_KEEPALIVE
+          # Close the subscriber connection cooperatively so the blocked read
+          # errors and the backend's rescue/retry reconnects on a fresh socket.
+          # Reset @last_message to avoid cascade-triggering during the 1s reconnect window.
+          @last_message = Time.now
+          backend_instance.request_reconnect
         end
+
+        timer.queue(keepalive_interval, &blk) if keepalive_interval > MIN_KEEPALIVE
       end
     end
 
--- a/lib/message_bus/backends/base.rb
+++ b/lib/message_bus/backends/base.rb
@@ -175,6 +175,12 @@ module MessageBus
         raise ConcreteClassMustImplementError
       end
 
+      # Closes the subscriber's connection so its blocked read returns an error
+      # and the backend's own rescue/retry re-establishes it cooperatively.
+      # Backends that cannot get a stuck connection may leave this as a no-op.
+      def request_reconnect
+      end
+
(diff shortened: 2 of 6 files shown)


Mubramaj:
Hi @SamSaffron , @jeremyevans 

Just a follow up on this. I have been running on production this branch since April and I had no side effect. Everything works as expected. Everytime I have the timeout it reconnects correctly. 

Before MessageBus would be dead and I had to manually restart the server, now it self heals without any issue or downtime. 

Happened again this morning: 

image

jeremyevans:
@Mubramaj I apologize, but I don't think I would have time to look into this in the near future.

Mubramaj:
No pb I will continue using my branch for now and keep it updated with the changes in the main repo

SamSaffron:
change looks reasonably modest, will test today and see if I can merge.

SamSaffron:
Looks like a couple of blockers here: 

1. A single exception during a reconnect attempt permanently stops the keepalive watchdog.
2. The Postgres implementation closes a PG::Connection from a different thread while the subscriber thread is using it, which
   violates libpq’s threading requirements and can potentially cause a native crash.

Mubramaj:
Thank you for the feedback I will work on it as soon as possible.

Mubramaj:
Ok @SamSaffron , 

Changes have been implemented:

1. Wrapped the keepalive proc's body in begin/ensure with the watchdog re-arming in the ensure so it never skips even if there is an exception in the reconnect

2. Use a request-reconnect flag instead of closing the connection directly from the watchdog thread. The subscriber thread checks this flag itself and reconnects on its own, so the connection is never touched by a thread other than the one that owns it. 

On top of it added some small fixes on the test setup that I encountered while testing using `docker-compose run tests`:

1. A Stale Dockerfile pinning ruby 2.7 while the gemspec requires ruby >= 3.2.0
2. In `helpers.rb` wire REDISURL through redis_config so Redis specs actually honor it instead of silently defaulting to localhost:6379 


Thanks for the valuable feedbacks

SamSaffron:
Getting there but a careful agent review did find that Postgres reconnect currently loses messages

you can see my full session here: https://gisthost.github.io/?28b247923b2ae0bfb3d62b450beea873/index.html 

Thanks for taking this! we are getting close.

Mubramaj:
Hi Sam,

Since there have been lots of feedback on Postgres backend I stepped back and started thinking: my initial issue was for Redis backend, and after analysis I realised that Postgres does not have the issue at all.

If we look at the Redis backend: `redis.rb line 294` `global_redis.subscribe` is a blocking call with no read timeout. The TCP connection going half open makes this call sit there forever without error or timeout so no retry or recovery.

```ruby
          global_redis.subscribe(redis_channel_name) do |on|
            on.subscribe do
              clear_backlog.call(&blk) if highest_id
              @subscribed = true
            end

            on.unsubscribe { @subscribed = false }

            on.message do |_c, m|
              # ...
            end
          end

```

For the Postgres it's different: `postgres.rb line 176` `conn.wait_for_notify(10)` is called inside the while loop. So even if there is a dead, or half-open socket this call will returns every 10 seconds. Worst case scenario we lose the data for 10 seconds.

```ruby
            while listening_on?(channel, obj)
              raise ReconnectRequested if reconnect_requested?

              conn.wait_for_notify(10) do |_, _, payload|
                break unless listening_on?(channel, obj)

                listener.do_message.call(nil, payload)
              end
            end
```

Do you think I should do nothing for Postgres with a nice comment that explains that, this way we don't have too much code change on this PR, or do you want me to keep going on the Postgres feedback to stay consistent on the reconnect between Redis and Postgres?

Mubramaj:
Hi Sam, 

I worked on all the feedback from the reviewer Agent. 

About the data loss, I think this was a pre-existing gap where postgres backend was never incrementing the `highest_id`. Maybe the postgres backend is not really used by the dev community ? 

Anyway it has been fixed in this PR. 


Side question: the reviewer agent session you sent me is pretty cool. Did you code it yourself ? What tech/stack did you use?

SamSaffron (approved):

SamSaffron:
Thanks so much for fixing, I think there is one more pre-existing issue, but we can sort it out in a diff commit . unsubscribe can be lost during reconnect. we can sort it in a followup ... it is not new.

for now though I am merging this, thanks so much!

regarding my workflows ... I have been working on https://term-llm.com/ in my spare time, this is the agent I use for all of my work including the reviews here. 

image

I tend to run a lot of llms and will run lots of group reviews which tend to surface many edge cases. At the moment I am focusing on the web surface, the tui surface is very mature.

Mubramaj:
You are welcome it's always rewarding contributing on discourse open source code. I learn a lot from the code and from the valuable feedback! 

Congrats on term-llm I will definitely be checking that out !

